PR #87, opened 2025-04-02
Feat: Modify T1Chain to allow 7683 proof of fill

kss-t1:
## Description
Modification of T1Chain to allow for Proof of Fill Linked List data.

## Motivation and Context
T1Chain needs to hold Proof of Fill data so that we can prove to L1 users that their Orders were Filled on t1

-   [ ] Docs

## Checklist:

-   [ ] This change includes any required test coverage.


contracts/src/L1/rollup/T1Chain.sol
```diff
@@ -531,7 +524,7 @@ contract T1Chain is OwnableUpgradeable, PausableUpgradeable, IT1Chain {
             BatchHeaderV0Codec.getL1MessagePopped(batchPtr)
         );
 
-        _afterFinalizeBatch(_totalL1MessagesPoppedOverall, _batchIndex, _batchHash, _postStateRoot, _withdrawRoot);
+        _afterFinalizeBatch(_totalL1MessagesPoppedOverall, _batchIndex, _batchHash, _postStateRoot, _withdrawRoot, "");
```
diego-G, line 527: I'm hesitant about this cause it is set to an empty array when using this method. Would that be expected?

contracts/src/L1/rollup/IT1Chain.sol
```diff
@@ -118,8 +122,16 @@ interface IT1Chain {
     /// @notice t1 batch finalization
     ///
     /// @param withdrawRoot The withdraw trie root of current batch.
-    /// @param signature The ECDSA valid signature to match one of the provers
-    function finalizeBatchWithProof(bytes32 withdrawRoot, bytes calldata signature) external;
+    /// @param withdrawRootSignature The ECDSA valid signature of withdraw root to match one of the provers
+    /// @param proofOfFill7683Root 7683 The Proof of Fill trie root of current batch.
+    /// @param proofOfFillRootSignature The ECDSA valid signature of proof of fill root to match one of the provers
+    function finalizeBatchWithProof(
```
diego-G, line 128: WDYT about overloading `finalizeBatchWithProof()` instead of modifying it? I haven't checked the contract size.

diego-G (changes requested):

evchip (dismissed):

contracts/src/L1/rollup/T1Chain.sol
kss-t1: We are not using this method so I am not going to modify its functionality. This is Scroll legacy.

contracts/src/L1/rollup/IT1Chain.sol
kss-t1: Why? It is intended that updates to both Tries are tied together.

evchip: I agree with Diego's suggestion, but let's not block this further.

evchip (approved):
ty ty ty

PR #92, opened 2025-04-11
Test: Replace DsTestPlus with Forge Std Test

evchip:
## Description
Replace antiquated DsTestPlus with far superior forge std library Test

## Motivation and Context
forge std library's Test has [more features](https://book.getfoundry.sh/cheatcodes/)

## How Has This Been Tested?
All tests are passing

### DIFF
--- a/contracts/script/test/SwapERC20.s.sol
+++ b/contracts/script/test/SwapERC20.s.sol
@@ -3,9 +3,6 @@
 pragma solidity ^0.8.25;
 
 import { Script } from "forge-std/Script.sol";
-import { StdUtils } from "forge-std/StdUtils.sol";
-
-import { DSTestPlus } from "solmate/test/utils/DSTestPlus.sol";
 
 import { ISignatureTransfer } from "@uniswap/permit2/src/interfaces/ISignatureTransfer.sol";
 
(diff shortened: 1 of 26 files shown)


contracts/script/test/SwapERC20.s.sol
```diff
@@ -3,9 +3,6 @@
 pragma solidity ^0.8.25;
 
 import { Script } from "forge-std/Script.sol";
-import { StdUtils } from "forge-std/StdUtils.sol";
```
diego-G, line 6: Please, check for other imports like this. We can remove them.

contracts/src/test/L1GasPriceOracle.t.sol
```diff
@@ -2,12 +2,12 @@
 
 pragma solidity ^0.8.25;
 
-import { DSTestPlus } from "solmate/test/utils/DSTestPlus.sol";
```
diego-G, line 5: We might be in a good position now to avoid using `solmate` dependency. Are we?

diego-G (dismissed):

contracts/src/test/L1GasPriceOracle.t.sol
evchip: we are using solmate's WETH and MockERC20 in a few places

diego-G (approved):

PR #98, opened 2025-04-26
Release: v0.2.2

evchip:
## Description
* `v0.2.1.json`: add metadata for upgrading `L1_T1_CHAIN`
* `v0.2.0.json`: add missing `L1_PROXY_ADMIN_ADDR`

## Motivation and Context
We had an issue with self claiming withdrawals on v0.2.0. The batch index is still hardcoded as 1 in the contract, which means that we continually override the same withdraw trie root every time `finalizeBatchWithProof` is called, which means that users will only have ~12 seconds to call `relayMessageWithProof` after the root that corresponds to their withdrawal is posted. This change upgraded the T1Chain contract so that we can circumvent this issue.

## How Has This Been Tested?
Tested that Postman is now able to process `finalizeBatchWithProof` by passing in batch index

## Checklist:


contracts/deployments/bridge/v0.2.2.json
```diff
@@ -0,0 +1,13 @@
+{
+  "metadata": {
+    "version": "v0.2.2",
+    "date": "25-04-2025",
+    "by": "Evan",
+    "from_machine": "Evan's Macbook",
+    "from_branch": "canary",
+    "from_tag": "v0.2.1"
+  },
+  "addresses": {
```
diego-G, line 11: Can we somehow write where the rest of addresses can be found?

diego-G (changes requested):

contracts/deployments/bridge/v0.2.2.json
evchip: done

diego-G (approved):

PR #99, opened 2025-04-28
Feat: Verify Proof Of Fill before 7683 settlement

diego-G:
## Description

Relay 7683 settlements from target chain including an intent filled Merkle proof.

## Motivation and Context

Currently, `T1Chain` L1 contract is holding a `proofOfFill7683Roots` which maps a finalised batch to a root of `ProofOfFill7683` Trie valid for that particular batch. 7683 handle interface has been modified so that client contracts can submit a proof of fill Merkle inclusion root.  It then validate the proofs using a specific `orderId`, generate a Merkle Root based on these validations and compare it with Proof of Fill Root saved in `T1Chain`.

## How Has This Been Tested?

`forge test --match-contract T1BasicSwapE2E`

## Checklist:

### DIFF
--- a/contracts/src/7683/T1ERC7683.sol
+++ b/contracts/src/7683/T1ERC7683.sol
@@ -64,14 +64,14 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable {
     /// @param _origin The origin domain
     /// @param _sender The sender address
     /// @param _batchIndex The index of the batch where the PoF belongs to
-    /// @param _proofOfFillMerkleProof The proof of fill Merkle root
+    /// @param _intentFilledMerkleProof The intent filled Merkle proof
     /// @param _nonce The nonce of the message
     /// @param _message The message
     function handle(
         uint32 _origin,
         bytes32 _sender,
         uint256 _batchIndex,
-        bytes calldata _proofOfFillMerkleProof,
+        bytes calldata _intentFilledMerkleProof,
         uint256 _nonce,
         bytes calldata _message
     )
@@ -79,7 +79,7 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable {
         payable
         onlyMessenger
     {
-        _handle(_origin, _sender, _batchIndex, _proofOfFillMerkleProof, _nonce, _message);
+        _handle(_origin, _sender, _batchIndex, _intentFilledMerkleProof, _nonce, _message);
     }
 
     // ============ Internal Functions ============
@@ -123,14 +123,14 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable {
     /// @param _messageOrigin The domain from which the message originates
     /// @param _messageSender The address of the sender on the origin domain
     /// @param _batchIndex The index of the batch where the PoF belongs to
-    /// @param _proofOfFillMerkleProof The proof of fill merkle root
+    /// @param _intentFilledMerkleProof The intent filled Merkle proof
     /// @param _nonce The nonce of the message
     /// @param _message The encoded message received via t1
     function _handle(
         uint32 _messageOrigin,
         bytes32 _messageSender,
         uint256 _batchIndex,
-        bytes calldata _proofOfFillMerkleProof,
+        bytes calldata _intentFilledMerkleProof,
         uint256 _nonce,
         bytes calldata _message
     )
@@ -146,10 +146,10 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable {
 
         if (
             !WithdrawTrieVerifier.verifyMerkleProof(
-                t1Chain.proofOfFill7683Roots(_batchIndex),
+                t1Chain.intentFilledMerkleProofs(_batchIndex),
                 keccak256(abi.encode(_settle, _orderIds, _ordersFillerData)),
                 _nonce,
-                _proofOfFillMerkleProof
+                _intentFilledMerkleProof
             )
         ) {
             revert InvalidProof();
--- a/contracts/src/7683/T1ERC7683.sol
+++ b/contracts/src/7683/T1ERC7683.sol
@@ -39,7 +39,6 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable {
 
     // ============ Errors ============
     error OnlyMessenger();
-    error FunctionNotImplemented(string functionName);
     error EthNotAllowed();
     error BatchNotFinalized();
     error IntentProofNotFound(uint256 batchIndex, bytes32 proofOfFillRoot);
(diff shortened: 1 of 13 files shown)


contracts/src/7683/T1ERC7683.sol
```diff
@@ -105,11 +123,46 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable {
...
+    {
+        IT1Chain t1Chain = IT1Chain(IL1T1Messenger(address(messenger)).rollup());
+
+        // Check if intent proof batch index is finalized
+        if (!t1Chain.isBatchFinalized(_batchIndex)) revert BatchNotFinalized();
+
+        bytes32 _proofOfFillRoot = t1Chain.proofOfFill7683Roots(_batchIndex);
+
+        // Check intent proof exists in our rollup
+        if (_proofOfFillRoot != _proofOfFillRoot) {
```
evchip: ?

contracts/src/7683/T1ERC7683.sol
```diff
@@ -105,11 +123,46 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable {
...
+            revert IntentProofNotFound(_batchIndex, _proofOfFillRoot);
```
evchip: To me this error message does not accurately describe why this error would be thrown

contracts/src/test/7683/T1BasicSwapE2E.t.sol
```diff
@@ -697,26 +774,38 @@ contract T1BasicSwapE2E is BaseTest {
...
 
-        bytes memory outerMessage =
-            abi.encodeWithSelector(T1ERC7683.handle.selector, destination, destinationRouterB32, innerMessage);
-
         // hash 0xcca132db240c06c148d210ceda18701a38e863e5ab2ed4638b15b6c7b30a08ae
         uint256 nonce = 0;
         uint256 msgValue = 0;
         address from = address(destinationRouter);
         address to = address(originRouter);
+        bytes memory proof = hex"";
```
evchip, line 837: could we do tests a couple of proofs at different heights? perhaps up to height 2

contracts/src/7683/T1ERC7683.sol
diego-G: This is a vestige from a previous version. Removed

contracts/src/7683/T1ERC7683.sol
```diff
@@ -60,9 +63,23 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable {
...
+    /// @param _proofOfFillMerkleProof The proof of fill Merkle root
+    /// @param _nonce The nonce of the message
     /// @param _message The message
-    function handle(uint32 _origin, bytes32 _sender, bytes calldata _message) external payable onlyMessenger {
-        _handle(_origin, _sender, _message);
+    function handle(
+        uint32 _origin,
+        bytes32 _sender,
+        uint256 _batchIndex,
+        bytes calldata _proofOfFillMerkleProof,
```
evchip: i think `_proofOfFillMerkleProof` is a bad name. how about `intentFilledMerkleProof`?

contracts/src/test/7683/T1BasicSwapE2E.t.sol
```diff
@@ -804,7 +844,7 @@ contract T1BasicSwapE2E is BaseTest {
             "relayMessage(address,address,uint256,uint256,bytes)", from, to, msgValue, nonce, outerMessage
         );
 
-        rollup.setProofOfFill7683Root(batchIndex, keccak256(innerMessage));
+        rollup.setProofOfFill7683Root(batchIndex, proofOfFill7683Root);
```
evchip: we could probably call `finalizeBatchWithProof` instead of `finalizeBundleWithProof` to get around using a custom setter

contracts/src/7683/T1ERC7683.sol
diego-G: Shall we change the [varname](https://github.com/t1protocol/t1/blob/345b6c5efa726f3bd639029249038f48012cfaf4/contracts/src/L1/rollup/T1Chain.sol#L170) that we save in T1Chain as well?

evchip:
per discussion we want to remove the modifier on `handle` so that it can be called directly by the solver (or anyone interested in getting the solver paid back).

Previous flow: triggering the settlement via the `settle` method on the target chain, triggering a `MessageSent` event picked up by the off chain relayer, which triggered a proof generation in the withdraw trie and calling of `relayMessageWithProof` which triggered settlement of intent.

New flow: when `fill` event is called on target chain it emits a `Filled` event which is picked up by off chain relayer, hashed, hash appended to intent-specific merkle trie, and inclusion proof generated. proof is relayed to solver via API (TODO) and solver calls `handle` (or whatever, could be called `settle` as well -- this is not part of the 7683 spec) on the source chain 7683 contract, passing in the intent filled inclusion proof, which releases the funds to the solver.

contracts/src/7683/T1ERC7683.sol
evchip: sounds good, thanks. i believe KSS renamed it to something better in postman so perhaps you can align on that name

kss-t1: This is still not renamed in Postman. It should be, but because it isn't , that should not stop us from doing that correctly  here.
I like `intentFilledMerkleProof` .

diego-G: [chore(7683): rename proofOfFill*-\>intentFilled*](https://github.com/t1protocol/t1/pull/99/commits/8359daf367b31e5fa707d8d7ed4c8493f2acc1b0)

kss-t1:
> per discussion we want to remove the modifier on `handle` so that it can be called directly by the solver (or anyone interested in getting the solver paid back).
> 
> Previous flow: triggering the settlement via the `settle` method on the target chain, triggering a `MessageSent` event picked up by the off chain relayer, which triggered a proof generation in the withdraw trie and calling of `relayMessageWithProof` which triggered settlement of intent.
> 
> New flow: when `fill` event is called on target chain it emits a `Filled` event which is picked up by off chain relayer, hashed, hash appended to intent-specific merkle trie, and inclusion proof generated. proof is relayed to solver via API (TODO) and solver calls `handle` (or whatever, could be called `settle` as well -- this is not part of the 7683 spec) on the source chain 7683 contract, passing in the intent filled inclusion proof, which releases the funds to the solver.

Agreed. `handle` / `settle` needs to be callable by `Filler` , or anyone else incentivised to unlock his funds for him.

diego-G:
Solution parked in favor xchain read primitive.

PR #105, opened 2025-05-05
Refactor: `DeployT1ERC7683Pull`

evchip:
## Description
* refactor(`DeployT1ERC7683Pull`): split into 4 contracts so that run method is default and it is easier to integrate into our existing deployment scripts
* docs(`.env.template`): add L1 and L2 signer needed for `xChainRead`

## Motivation and Context
@w-woloszyn had some trouble deploying these contracts because they use a different flow compared to the rest of the deployment scripts in `deploy.sh`

## How Has This Been Tested?
I ran `deploy.sh`

## Checklist:

-   [x] I have added/updated any deployment scripts to ensure that the protocol is functional (Makefile, Dockerfile, Forge Script, etc.).


contracts/.env.template
```diff
@@ -21,6 +21,8 @@ L1_SECURITY_COUNCIL_PRIVATE_KEY=
 TEST_PRIVATE_KEY=
 ALICE_PRIVATE_KEY=
 MARKET_MAKER_PRIVATE_KEY=
+L1_SIGNER= # same as postman L1 signer
```
w-woloszyn, line 26: Oof! Thanks for catching that. I think I missed it in my PR

contracts/deploy.sh
```diff
@@ -84,6 +84,10 @@ run_script script/deploy/InitializeL2BridgeContracts.s.sol InitializeL2BridgeCon
 run_script script/deploy/InitializeL1T1Owner.s.sol InitializeL1T1Owner
 run_script script/deploy/DeployL2T1XChainReader.s.sol DeployL2T1XChainReader
 run_script script/deploy/DeployL1T1XChainReader.s.sol DeployL1T1XChainReader
+run_script script/deploy/7683/DeployT1ERC7683Pull.s.sol DeployT1ERC7683PullL1
```
w-woloszyn, line 95: Have you tested it? I don't think it is going to work. Currently the script is doing a glob search for `L1` and `L2` to decide where to deploy, as described in https://www.notion.so/t1protocol/refactor-contract-deployment-1e8231194dc380d293d5f7d3ee3a39e5

contracts/script/README.md
```diff
@@ -34,6 +34,11 @@ The magical deploy order is as follows:
 16. [DeployL2T1XChainReader.s.sol](./deploy/DeployL2T1XChainReader.s.sol)
 17. [DeployT1ERC7683Pull.s.sol](./deploy/7683/DeployT1ERC7683Pull.s.sol)
```
w-woloszyn: Is this still relevant? We have to do that before doing the bulletpoint steps? Also, I think add it to the numbered list and if this is not a part of usual deployment then we change it. My point is that release process should be extremely easy so that a monkey can follow it. Right now, the difference numbered/bulletpoints can raise eyebrows (do monkeys have eyebrows?)

w-woloszyn (changes requested):

w-woloszyn:
@evchip gentle bump

contracts/script/README.md
evchip: I think our deployment process will change quite a bit in the next few iterations so this whole readme is probably going to be outdated. Let's accept that there will be some growing pains and I will help to patch the holes as we see them.

(TIL yes, monkeys do have eyebrows!)

contracts/deploy.sh
evchip: I think you are right. Unfortunately this is non trivial to fix. I'm thinking to abandon this PR until we have more info on our needs for the next deployment.

w-woloszyn:
@evchip gentler bump

evchip:
@w-woloszyn I'm thinking to close this. Are you cool with that?

w-woloszyn:
Sure!

PR #106, opened 2025-05-06
Refactor: handle xCR callback failure

evchip:
## Description
* refactor `BaseT1XChainReader.sol`: add try/catch to `onT1XChainReaderResult` call
* test `T1XChainReaderTest`: add tests for `ReadFailed` reverts + refactor test setups to be DRY

## Motivation and Context
To gracefully handle errors originating from the callback contract. The benefit of this is approach is that the error is visible and overt for the user. Reverting at the top level means that the callback would fail at the gas estimation step, meaning the user never even sees the result of their xCR request delivered, and has no information about why it failed. With this try/catch, the user will see the reason.

## How Has This Been Tested?
Added 2 tests

## Checklist:

-   [x] I have added/updated any deployment scripts to ensure that the protocol is functional (Makefile, Dockerfile, Forge Script, etc.).

### DIFF
--- a/contracts/src/libraries/xChain/BaseT1XChainReader.sol
+++ b/contracts/src/libraries/xChain/BaseT1XChainReader.sol
@@ -59,8 +59,8 @@ abstract contract BaseT1XChainReader is OwnableUpgradeable, ReentrancyGuardUpgra
 
     // ============ State Variables ============
 
-    IT1Messenger public immutable messenger;
     /// @notice The T1 messenger contract used for cross-chain communication
+    IT1Messenger public immutable messenger;
 
     /// @notice The T1 prover
     address public immutable prover;
--- a/contracts/src/libraries/xChain/BaseT1XChainReader.sol
+++ b/contracts/src/libraries/xChain/BaseT1XChainReader.sol
@@ -213,6 +213,5 @@ abstract contract BaseT1XChainReader is OwnableUpgradeable, ReentrancyGuardUpgra
                 emit ReadFailed(requestId, result, reason);
             }
         }
-
     }
 }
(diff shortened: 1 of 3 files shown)


contracts/src/libraries/xChain/BaseT1XChainReader.sol
```diff
@@ -189,10 +205,13 @@ abstract contract BaseT1XChainReader is OwnableUpgradeable, ReentrancyGuardUpgra
         // If there's a valid callback, forward the result
         if (callback != address(0)) {
             delete callbacks[requestId];
-
-            IT1XChainReaderCallback(callback).onT1XChainReaderResult(_originDomain, _sender, requestId, result);
+            try IT1XChainReaderCallback(callback).onT1XChainReaderResult(_originDomain, _sender, requestId, result) {
+                emit ReadSucceeded(requestId, result);
```
diego-G, line 209: Missing test for the successful event.

contracts/src/libraries/xChain/BaseT1XChainReader.sol
```diff
@@ -189,10 +205,13 @@ abstract contract BaseT1XChainReader is OwnableUpgradeable, ReentrancyGuardUpgra
...
+            } catch Error(string memory reason) {
+                emit ReadFailed(requestId, result, reason);
```
diego-G, line 211: This event is not being tested either.

diego-G (changes requested):
Two missing tests. The rest LGTM

contracts/src/libraries/xChain/BaseT1XChainReader.sol
evchip: done

contracts/src/libraries/xChain/BaseT1XChainReader.sol
evchip: done

diego-G (approved):

PR #113, opened 2025-05-26
Feat/emit xchain read request initiatior

kss-t1:
## Description

Emits address which initiated x-chain request in the `ReadRequested` event.

## Motivation and Context

This is needed so that users can filter out their own Merkle Proofs when requesting x-chain Merkle Proofs from the API.

-   [ ] Docs

## Checklist:

-   [ ] This change includes any required test coverage.

### DIFF
--- a/contracts/src/libraries/xChain/T1XChainReader.sol
+++ b/contracts/src/libraries/xChain/T1XChainReader.sol
@@ -21,6 +21,7 @@ contract T1XChainReader is OwnableUpgradeable, ReentrancyGuardUpgradeable {
      * @param requestId Unique identifier for the request
      * @param destinationDomain Domain ID of the target chain
      * @param targetContract Address of the contract to read from
+     * @param requester Address who initiated the read request
      * @param gasLimit The gas limit for the read operation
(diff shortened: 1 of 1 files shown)


contracts/src/libraries/xChain/T1XChainReader.sol
```diff
@@ -30,6 +31,7 @@ contract T1XChainReader is OwnableUpgradeable, ReentrancyGuardUpgradeable {
         bytes32 indexed requestId,
         uint32 indexed destinationDomain,
         address targetContract,
+        address requester,
```
evchip, line 34: why not `requestooor`?

evchip (approved):

contracts/src/libraries/xChain/T1XChainReader.sol
kss-t1: you will laugh but I actually looked it up BEFORE making that decision. and I landed with `requester` because AI told me this: 
```
"Requestor" is a less common spelling of "requester," both referring to the person who makes a request. 
```

PR #116, opened 2025-05-30
feat: switch settle verification to merkle proof

aharvet:
## Description

Change the callback mechanism for filler repayment to a Merkle Trie of xChain Reads , and make it necessary for the Filler to perform an Inclusion Proof in order to unlock his funds on the source chain.

## Motivation and context

We do want the PoR Merkle Trie because:

- That way we do not subsidize L1 gas
- This allows other parties (such as dapp devs) to submit the proof and therefore unlock the funds. With depending on TEE entity submitting the xChain read callback instead, that can only be done by TEE EOAs owned by t1

## How Has This Been Tested?

Updated previous tests to match changes.
Added tests for the various scenarios introduced by the merkle tree.
There is room for completing the test suite with other scenarios depending on how far we want to go with it.

-   [x] New feature (breaking change which adds functionality)

## Checklist:


contracts/src/7683/T1ERC7683Pull.sol
```diff
@@ -39,6 +44,10 @@ contract T1ERC7683Pull is BasicSwap7683, OwnableUpgradeable, IT1XChainReaderCall
      * @param isSettled Whether the order is settled
      */
     event SettlementVerified(bytes32 indexed orderId, bool isSettled);
+    /**
+     * @notice Emitted when a proof of read root is committed
+     */
+    event ProofOfReadRootCommitted(uint256 batchIndex);
```
evchip: could add natspec comment for the param too

contracts/src/7683/T1ERC7683Pull.sol
```diff
@@ -109,25 +119,36 @@ contract T1ERC7683Pull is BasicSwap7683, OwnableUpgradeable, IT1XChainReaderCall
...
-    /// @param requestId The ID of the read request
-    /// @param result The result of the read
-    function onT1XChainReaderResult(
-        uint32 _originDomain,
-        bytes32 _sender,
+    // @notice Callback function for cross-chain read results
+    // @param _originDomain The origin domain
+    // @param _sender The sender address
+    // @param requestId The ID of the read request
+    // @param result The result of the read
```
evchip: this is no longer correct

contracts/src/libraries/xChain/T1XChainReader.sol
```diff
@@ -191,53 +189,35 @@ contract T1XChainReader is OwnableUpgradeable, ReentrancyGuardUpgradeable {
...
-        bytes32 _requestId,
-        bytes memory _data
-    )
-        external
-        payable
-        onlyProver
-    {
-        _handleReadResponse(_originDomain, _sender, _requestId, _data);
+    function handle(bytes32 requestId, uint256 batchIndex, bytes32 newRoot) external payable onlyProver {
+        _handleReadResponse(requestId, batchIndex, newRoot);
```
evchip: IMO it would be better to store `nextBatchIndex` on-chain and increment it similar to the `T1Chain` contract.

evchip (changes requested):
I envisioned this differently: the roots are stored in the `T1XChainReader` contract, but any contract that uses `T1XChainReader` will not need to fetch the root, they just call some method like `T1XChainReader.verify(bytes32 requestId, bytes calldata result, bytes calldata proof)` which under the hood calls `WithdrawTrieVerifier.verifyMerkleProof`. That way we encapsulate the proving logic (perhaps later we want to change the data structure to an [IAVL tree](https://github.com/cosmos/iavl)), and ensure that any user of xChainReader doesn't have to manage batch indices and merkle roots.

evchip (commented):
7683 doesn't have to pass `ReadRequest.callback` to `requestRead` anymore

contracts/src/7683/T1ERC7683Pull.sol
```diff
@@ -21,6 +24,8 @@ contract T1ERC7683Pull is BasicSwap7683, OwnableUpgradeable, IT1XChainReaderCall
     address public counterpart;
 
     // ============ State Variables ============
+    mapping(uint256 batchIndex => bytes32 root) public proofOfReadRoots;
```
kss-t1: This should be stored in `T1XChainReader` . `T1ERC7683Pull` is just a client contranct to the Proof of Read Merkle Trie logic.

contracts/src/7683/T1ERC7683Pull.sol
```diff
@@ -121,31 +106,22 @@ contract T1ERC7683Pull is BasicSwap7683, OwnableUpgradeable, IT1XChainReaderCall
...
         bytes calldata proof
     )
         external
     {
-        bytes32 xChainReadResultHash = keccak256(result);
-        bytes32 leaf = keccak256(abi.encodePacked(xChainReadResultHash, requestId));
-        uint256 batchIndex = requestIdToBatchIndex[requestId];
-
-        if (!WithdrawTrieVerifier.verifyMerkleProof(proofOfReadRoots[batchIndex], leaf, nonce, proof)) {
+        if (!xChainRead.verifyProofOfRead(batchIndex, requestId, position, result, proof)) {
```
evchip: contrary to my initial suggestion, I think we should pack all fields into a single `bytes` param to allow us maximum flexibility in later reconfiguring the params without affecting any consumers of `xChainRead`

contracts/src/libraries/xChain/T1XChainReader.sol
```diff
@@ -190,55 +185,31 @@ contract T1XChainReader is OwnableUpgradeable, ReentrancyGuardUpgradeable {
...
-        bytes32 _sender,
-        bytes32 _requestId,
-        bytes memory _data
-    )
-        external
-        payable
-        onlyProver
-    {
-        _handleReadResponse(_originDomain, _sender, _requestId, _data);
+    function commitProofOfReadRoot(uint256 batchIndex, bytes32 newRoot) external payable onlyProver {
```
kss-t1, line 166: Can we please track `nextBatchIndex` as a public field of this contract?
It would hold batch index which `prover` should use if he wants to add new batch of the Trie updates.
`batchIndex` should still remain in `commitProofOfReadRoot` so that `prover` can overwrite past trie roots.
If `prover` tries to submit a `batchIndex` higher than `nextBatchIndex`, the transaction should revert.

kss-t1 (dismissed):

contracts/src/libraries/xChain/T1XChainReader.sol
```diff
@@ -190,55 +188,35 @@ contract T1XChainReader is OwnableUpgradeable, ReentrancyGuardUpgradeable {
...
+        if (batchIndex > nextBatchIndex) revert InvalidBatchIndex();
```
evchip, line 167: looks like `nextBatchIndex` is not incremented anywhere

contracts/src/libraries/xChain/T1XChainReader.sol
```diff
@@ -35,38 +35,43 @@ contract T1XChainReader is OwnableUpgradeable, ReentrancyGuardUpgradeable {
...
 
     /**
      * @notice Emitted when a cross-chain read response fails with string revert
      * @param requestId Unique identifier for the original request
-     * @param result The result data from the read operation
+     * @param batchIndex The batch index of the read request
      * @param reason The reason the call failed
      */
-    event ReadFailed(bytes32 indexed requestId, bytes result, string reason);
+    event ReadFailed(bytes32 indexed requestId, uint256 batchIndex, string reason);
```
evchip: Looks like these aren't being used anymore

contracts/src/test/7683/T1XChainReaderTest.t.sol
```diff
@@ -188,4 +273,76 @@ contract T1XChainReaderTest is T1BasicSwapE2E {
...
         return (orderData, orderId_, requestId_);
     }
+
+    /// @dev Generate a merkle tree with depth 2 (4 leaves) including the result value and a proof
+    /// @param requestId Proof of read request id
+    /// @param result Result of the read
+    /// @param position position of the leaf where result is stored (0-3)
+    /// @return root Root of the merkle tree
+    /// @return proof Proof of for the leaf where result is stored
+    function _generateMerkleTree(
```
evchip, line 305: this is cooool

evchip (dismissed):
Looks good, thanks!

aharvet:
Need one last approval because of rebase commit

evchip (dismissed):

evchip (approved):

PR #122, opened 2025-06-09
feat: emit ReadRequested event in T1XChainReader

w-woloszyn:
This change removes the dependency on IT1Messenger and instead directly emits a ReadRequested event from T1XChainReader. This enables off-chain listeners (e.g., the postman relayer) to observe read requests and fulfill them trustlessly by submitting results later using proofs.

## Types of changes (remove all unchecked types)

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Style (style only changes)
-   [ ] Docs
-   [ ] Refactor (code that does not add new functionality nor fixes a bug)

## Checklist:

-   [ ] If my change requires a change to the documentation, I have updated the documentation accordingly, and in either case, have checked this box to attest to my assessment of this requirement with regard to my change.
-   [ ] If my change requires additions or updates to any deployment scripts to ensure that the protocol is functional (Makefile, Dockerfile, Forge Script, etc.), I have made these changes, and in either case, have checked this box to attest to my assessment of this requirement with regard to my change.
-   [ ] If my change requires additional test coverage, I have created those tests accordingly, and in either case, have checked this box to attest to my assessment of this requirement with regard to my change.

PR #124, opened 2025-06-09
Release: v0.5.0 + xChainReader fixes

evchip:
## Description
- `T1ERC7683`: add tuple decoding to handle offset and length prefix added by off-chain relayer when executing xChainRead
- add `requester` field to `ReadRequest` struct this allows the off chain relayer to index proofs based on the sender of the request. otherwise there is no way to easily attribute the xCR requests
- add better natspec about ABI decoding on `verifyProofOfRead`
- Adds the metadata from my v0.5 contract (re)deployment
- Miscellaneous cleanup changes

## How Has This Been Tested?
tested against a new Postman instance on v0.5, with a [successful tx](https://sepolia.basescan.org/tx/0x082897c1ca0a98eacc9845dbcdc00003eb7a202d392a026282ad79d3d53d466e).

You can test it yourself:

1. Run one of the following:
```
// open USDT intent from Base to Arbitrum
forge script ./script/test/7683/T1ERC7683BaseToArb.s.sol:AliceSetupScript --broadcast
// open USDT intent from Arbitrum to Base
forge script ./script/test/7683/T1ERC7683ArbToBase.s.sol:AliceSetupScript --broadcast
```

2. go to `https://api.v05.t1protocol.com/api/read-proofs?address=<your address>&page=1&page_size=100`

3. copy the value in `handle_read_result_with_proof_calldata`

4. go to the block explorer (or use cast) and execute the [handleReadResultWithProof](https://sepolia.basescan.org/address/0xfc39ed3734f8660c96ea0523353a1cfd242e6c7d#writeContract#F2), passing in the value from step 3

-   [x] Release
-   [x] Fix

## Checklist:

### DIFF
--- a/contracts/src/7683/T1ERC7683.sol
+++ b/contracts/src/7683/T1ERC7683.sol
@@ -2,19 +2,26 @@
 pragma solidity ^0.8.25;
 
 import { OwnableUpgradeable } from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
-import { Hyperlane7683Message } from "../libraries/7683/Hyperlane7683Message.sol";
-import { BasicSwap7683 } from "./BasicSwap7683.sol";
-import { OrderData, OrderEncoder } from "../libraries/7683/OrderEncoder.sol";
+import { PausableUpgradeable } from "@openzeppelin/contracts-upgradeable/security/PausableUpgradeable.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { TypeCasts } from "@hyperlane-xyz/libs/TypeCasts.sol";
 
+import { GaslessCrossChainOrder, ResolvedCrossChainOrder, OnchainCrossChainOrder } from "../interfaces/IERC7683.sol";
+import { BasicSwap7683 } from "./BasicSwap7683.sol";
+import { Hyperlane7683Message } from "../libraries/7683/Hyperlane7683Message.sol";
+import { OrderData, OrderEncoder } from "../libraries/7683/OrderEncoder.sol";
 import { T1XChainReader } from "../libraries/xChain/T1XChainReader.sol";
+
 /**
  * @title T1ERC7683
  * @author t1 Labs
  * @notice This contract extends BasicSwap7683 with pull-based settlement using t1 cross-chain reads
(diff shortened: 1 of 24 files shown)


contracts/src/libraries/xChain/T1XChainReader.sol
```diff
@@ -113,29 +113,30 @@ contract T1XChainReader is OwnableUpgradeable, ReentrancyGuardUpgradeable {
         address targetContract,
         uint256 gasLimit,
         uint64 minBlock,
-        bytes calldata callData
+        bytes calldata callData,
+        address requester
```
aharvet, line 122: Is this change to allow anyone to request a proof for someone else ?

evchip: This is so that we can easily attribute proofs to the account that made the request. We want solvers to be able to pass their address into this API endpoint and get back any pending xCRs. Right now there is no other easy way to get this info, because we listen for the `SentMessage` event coming off the messenger contract, and that `msg.sender` is attributed to the xChainReader contract, rather than the user who called the contract. We can actually remove this param once we merge in [Wojciech's PR](https://github.com/t1protocol/t1/pull/122) to factor out the Messenger contract from the xCR flow.

evchip:
Note! This 7683 contract was deployed with the `orderId` param of the event `Filled` as `indexed`, even though this change is not represented in the diff here. Reason for no diff is that the library is imported as a dependency and we cannot override events in contracts. ref corresponding change in the [Postman PR](https://github.com/t1protocol/postman/pull/49/files#diff-fbe54c36a9d37740d7b388277fac6c0d0c0bb2f730f0e04337b65120daac8837R495). cc @kss-t1 @aharvet @princet1protocol

contracts/src/7683/T1ERC7683.sol
```diff
@@ -94,7 +94,8 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable {
             targetContract: counterpart,
             gasLimit: gasLimit,
             minBlock: 0,
-            callData: callData
+            callData: callData,
+            requester: _msgSender()
```
aharvet: I would rather use msg.sender for saving gas, as we'll very likely never use eip 2771 for meta transaction.

aharvet (dismissed):

contracts/src/7683/T1ERC7683.sol
evchip: fixed, thanks

aharvet (approved):

PR #125, opened 2025-06-10
feat: implement pausable pattern for 7683

aharvet:
## Description

Implemented a pausable pattern on `open` and `openFor` function to allow for blocking any new intent open.

## Motivation and Context

We want to assume that our Intent Bridges could be exploited and the funds locked in 7683 Escrows could be drained. If that happens, we will repay the users that lost their funds. We are however only willing to do that up to a threshold.

## How Has This Been Tested?

Introduced a `TvlThresholdTest.sol` test contract to specifically test the behavior of those functions depending on the pause state.

-   [ x ] New feature (non-breaking change which adds functionality)

## Checklist:

-   [ ] If my change requires a change to the documentation, I have updated the documentation accordingly, and in either case, have checked this box to attest to my assessment of this requirement with regard to my change.

### DIFF
--- a/contracts/src/7683/T1ERC7683.sol
+++ b/contracts/src/7683/T1ERC7683.sol
@@ -2,9 +2,18 @@
 pragma solidity ^0.8.25;
 
 import { OwnableUpgradeable } from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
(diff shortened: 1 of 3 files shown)


aharvet:
Should I document about this feature or is it straightforward enough ?

evchip (approved):
thanks!

PR #126, opened 2025-06-12
feat: implement upgrade script for chain reader

aharvet:
## Description

Add scripts to run an upgrade of the chain reader contract on sepolia.

Also adds scripts for local tests.

## Motivation and Context

The xChainReader contract is upgradeable but we don’t currently have a script to upgrade it. In fact we needed this recently in order to update the prover on the contract to avoid a signer clash on Postman, but were unable to quickly do so. Having a ready upgrade script would have helped us there.

## How Has This Been Tested?

Ran a deployment and then upgrade on a local network.

-   [x] Docs

## Checklist:

### DIFF
--- a/contracts/script/README.md
+++ b/contracts/script/README.md
@@ -110,8 +110,20 @@ Miscellaneous Scripts:
 Scripts to upgrade contract implementations:
 
 - [T1Chain implementation](./upgrade/UpgradeT1Chain.s.sol)
+
   ```bash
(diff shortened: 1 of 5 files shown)


evchip (approved):
LGTM. I wasn't able to verify the result of simulating an upgrade on Base Sepolia, but it seems to be a permissions issue, not an issue with the script itself.

PR #133, opened 2025-07-09
Feat/refund

aharvet:
## Description

Implement the refund functions to let users get their funds back in case their intent is not fill on time.

## Motivation and Context

So far, funds were locked on t1 contract once filldeadline was passed. We need to handle this situation.

## How Has This Been Tested?

Implemented a full testing suite for the new feature.

## Checklist:


contracts/src/7683/T1ERC7683.sol
```diff
@@ -202,46 +337,33 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable, PausableUpgradeable {
...
+    function _verifyOrderNotFilled(bytes32 orderId, bytes calldata encodedProofOfRead) internal view {
+        (bytes32 requestId, bytes memory result) = xChainRead.verifyProofOfRead(encodedProofOfRead);
 
-    /// @notice Not implemented
-    function _dispatchSettle(uint32, bytes32[] memory, bytes[] memory) internal pure override {
-        revert FunctionNotImplemented("_dispatchSettle");
-    }
+        // Verify this proof corresponds to the correct order
+        bytes32 expectedOrderId = readRequestToOrderId[requestId];
+        if (expectedOrderId != orderId) revert InvalidOrderStatus();
```
evchip: This feels like the wrong custom error type

contracts/src/7683/T1ERC7683.sol
```diff
@@ -127,7 +150,25 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable, PausableUpgradeable {
     /// @param destinationDomain The domain of the destination chain
     /// @param orderId The ID of the order to verify
     /// @return requestId The ID of the read request
-    function verifySettlement(uint32 destinationDomain, bytes32 orderId) external returns (bytes32 requestId) {
+    function verifySettlement(uint32 destinationDomain, bytes32 orderId) external payable returns (bytes32 requestId) {
+        requestId = _verifyFill(destinationDomain, orderId);
+        emit SettlementVerificationRequested(orderId, requestId);
+    }
+
+    function verifyRefund(uint32 destinationDomain, bytes32 orderId) external returns (bytes32 requestId) {
```
evchip: This is a potential DoS vector. Anyone can call `verifyRefund`, which will set `orderStatus[orderId] = REFUND_REQUESTED;`. After that it would be impossible to call `verifySettlement` because `orderStatus[orderId] != OPENED`, and there would be no way to change the status back to `OPENED`. This would permanently lock all user escrowed funds in the contract.

evchip (changes requested):
I found one critical issue on first pass. I'm not finished with the review but want to give you early feedback to address it.

contracts/src/7683/T1ERC7683.sol
aharvet: There was a logic behind it but I agree it's too stretched out. I'll change it

contracts/src/7683/T1ERC7683.sol
aharvet: Oh yes good catch

contracts/src/7683/T1ERC7683.sol
```diff
@@ -202,46 +337,33 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable, PausableUpgradeable {
...
+     */
+    function _verifyOrderNotFilled(bytes32 orderId, bytes calldata encodedProofOfRead) internal view {
+        (bytes32 requestId, bytes memory result) = xChainRead.verifyProofOfRead(encodedProofOfRead);
 
-    /// @notice Not implemented
-    function _dispatchSettle(uint32, bytes32[] memory, bytes[] memory) internal pure override {
-        revert FunctionNotImplemented("_dispatchSettle");
-    }
+        // Verify this proof corresponds to the correct order
+        bytes32 expectedOrderId = readRequestToOrderId[requestId];
```
evchip: I found another DoS vector. If someone requests a refund, an attacker could see this request and call `handleReadResultWithProof` after the prover commits the root but before the user can call `refund`. Since the proof is valid but has an empty `result` (meaning intent was not filled) then it will delete the `readRequestToOrderId[requestId]` entry but not settle the intent. At that point, when the user calls `refund`, `readRequestToOrderId[requestId]` will be empty and therefore `if (expectedOrderId != orderId) revert InvalidOrderStatus();` will be invoked.

contracts/src/7683/T1ERC7683.sol
```diff
@@ -127,7 +150,25 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable, PausableUpgradeable {
...
+        requestId = _verifyFill(destinationDomain, orderId);
+        emit SettlementVerificationRequested(orderId, requestId);
+    }
+
+    function verifyRefund(uint32 destinationDomain, bytes32 orderId) external returns (bytes32 requestId) {
+        (, bytes memory _orderData) = abi.decode(openOrders[orderId], (bytes32, bytes));
+        OrderData memory orderData = OrderEncoder.decode(_orderData);
+
+        if (orderData.originDomain != localDomain) revert InvalidOrderDomain();
+        if (block.timestamp <= orderData.fillDeadline) revert OrderFillNotExpired();
```
evchip: It seems that `fillDeadline` is for the destination chain, since it is checked in `_fillOrder`. So I don't think this would work.

contracts/src/7683/T1ERC7683.sol
evchip: There is another similar DoS issue here. If the user (or an attacker) calls `verifyRefund` but the intent actually was filled, the user can't get a refund (order was filled) and the solver can't settle the intent (wrong status). So even if there is a valid proof that the intent was filled, the solver couldn't redeem it because `orderStatus[orderId] != OPENED`

contracts/src/7683/T1ERC7683.sol
aharvet: Can you details why it wouldn't work ? to me the timestamp is linked to the intent not the chain. So we can check it's validity anywhere

evchip: Hm, it seems the exploit I was thinking of has been [patched since the Merge](https://ethereum.stackexchange.com/questions/135445/miner-modifiability-of-block-timestamp-after-the-merge). Timestamps did have some variation across blockchains. Miners could manipulate timestamps. It seems this isn't the case anymore, but I will still write out what my concern was, in case it's useful for thinking about edge cases:

Imagine the destination chain is Ethereum and the origin chain is Arbitrum. My concern was something like this:
1. user opens intent with short `fillDeadline`
2. user calls `verifyRefund`, and has the ability to slightly manipulate `fillDeadline` by a few seconds so that it has passed "early"
3. xChainRead proof of unfilled intent is generated
4. solver fills intent before actual deadline
5. if solver doesn't generate their own fill proof and call `handleReadResultWithProof` before `fillDeadline` has passed, the user can call `refund` which will be valid, and gets their initial funds back and gets their intent filled

Again, seems that this would no longer be possible.

aharvet: TLDR: let's add a delay after fill deadline to be extra secure.

Ok I understand what you meant.
So my initial reflection about this was that these kind of timestamp manipulations are bounded by the block time (+/- 12 seconds on ethereum for example).
And for now, we'll only launch on Arbitrum and Base that have low block frequency, 2.25 sec delta combined.
So the full workflow of PoR would probably be too long to be affected by such a manipulation.
Also considering those rollups are on a "trustmebro" model, I think it's reasonable to trust they won't manipulate the timestamps.

And the next chain where we are going to launch will probably have similar properties.

But it's true that we should be more careful with some chains that we could potentially integrate later.
So in order to be future proof and avoid an extra upgrade, maybe indeed it would be relevant to add an extra delay on top of fill deadline to make sure nothing can happen ?
Can also help for reorgs.
And I think adding 30 seconds for refunds specifically are not degrading UX too much.

evchip: Good points. Yes, I agree that a 30 second buffer should be a good safety precaution. Let's do it.

contracts/src/7683/BasicSwap7683.sol
```diff
@@ -9,6 +9,8 @@ import { TypeCasts } from "@hyperlane-xyz/libs/TypeCasts.sol";
 import { Base7683 } from "./Base7683.sol";
 import { OrderData, OrderEncoder } from "../libraries/7683/OrderEncoder.sol";
 
+import { console } from "forge-std/Test.sol";
```
evchip: rm please :)

contracts/src/7683/T1ERC7683.sol
```diff
@@ -180,6 +224,103 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable, PausableUpgradeable {
...
+            bytes32 orderId = _getOrderId(_orders[i]);
+            _verifyOrderNotFilled(orderId, _proofs[i]);
+            orderIds[i] = orderId;
+        }
+
+        _refundOrders(OrderEncoder.decode(_orders[0].orderData).originDomain, orderIds);
+    }
+
+    /**
+     * @dev Should be implemented by the messaging layer for dispatching a refunding instruction the remote domain
```
evchip: This description feels wrong. We are not dispatching a refund instruction, we are executing the refund.

contracts/src/7683/T1ERC7683.sol
```diff
@@ -180,6 +224,103 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable, PausableUpgradeable {
...
+     * where the orders where created.
+     * @param _originDomain The origin domain of the orders.
+     * @param _orderIds The IDs of the orders to refund.
+     */
+    /// @notice
+    /// @param originDomain The chain id of the network where intent has been created
+    /// @param orderIds Ids for the orders to refund
```
evchip: doubled natspec?

contracts/src/7683/T1ERC7683.sol
```diff
@@ -180,6 +224,103 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable, PausableUpgradeable {
...
+            _verifyOrderNotFilled(orderId, _proofs[i]);
+            orderIds[i] = orderId;
+        }
+
+        _refundOrders(OrderEncoder.decode(_orders[0].orderData).originDomain, orderIds);
+    }
+
+    /**
+     * @notice Refunds a batch of expired OnchainCrossChainOrder on the chain where the orders were opened.
+     * The refunded status should not be changed here but rather on the origin chain. To allow the user to retry in
```
evchip: "here" is ambiguous. It makes it sound as if the contract knows where it has been deployed.

contracts/src/7683/T1ERC7683.sol
```diff
@@ -180,6 +224,103 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable, PausableUpgradeable {
...
+    /// @notice Retrieves the local domain identifier.
+    /// @dev This function overrides the `_localDomain` function from the parent contract.
+    /// @return The local domain ID.
+    function _localDomain() internal view override returns (uint32) {
+        return localDomain;
+    }
+
+    /**
+     * @notice Refunds a batch of expired GaslessCrossChainOrders on the chain where the orders were opened.
+     * The refunded status should not be changed here but rather on the origin chain. To allow the user to retry in
```
evchip: same issue with "here"

contracts/src/7683/T1ERC7683.sol
```diff
@@ -127,7 +155,27 @@ contract T1ERC7683 is BasicSwap7683, OwnableUpgradeable, PausableUpgradeable {
     /// @param destinationDomain The domain of the destination chain
     /// @param orderId The ID of the order to verify
     /// @return requestId The ID of the read request
-    function verifySettlement(uint32 destinationDomain, bytes32 orderId) external returns (bytes32 requestId) {
+    function verifySettlement(uint32 destinationDomain, bytes32 orderId) external payable returns (bytes32 requestId) {
```
evchip, line 158: Good call on removing the `destinationDomain` param on `verifyRefund`. perhaps we should also use destination domain from `orderData` instead of passing it in?

aharvet: I thought of it but the increase in gas would be big here as we don't already get orderData like in verifyRefund.
Do you think it's still worth it ?

evchip: Ah I see. Good point, let's not do

evchip (approved):
Looks good to me, thanks!

PR #138, opened 2025-07-21
Feat/sealed bid auction/app structure

kss-t1:
## Description

Structure and basic logic of Sealed Bid TS App. This is supposed to run inside of a TEE to preserve absolute bid privacy.
PR includes:
- Basic `Bun` app structure
- `Winston` logging
- `Express` API server with startup and graceful shutdown
- Dummy controller

## Motivation and Context

We need this app to help determine 7683 Intent Solver and price for the Intent resolution

## How Has This Been Tested?

It was tested on local env.

-   [ ] Docs

## Checklist:

### DIFF
--- a/tee-aps/sealed-bid-auction/src/api/controller/AuctionController.ts
+++ b/tee-aps/sealed-bid-auction/src/api/controller/AuctionController.ts
@@ -1,0 +1,8 @@
+import type {Request, Response} from "express";
+
+export class AuctionController {
+
+    async open(req: Request, res: Response): Promise<void> {
+        res.status(501);
+    }
+}
--- a/tee-aps/sealed-bid-auction/src/core/SealedBidAuctionApiServer.ts
+++ b/tee-aps/sealed-bid-auction/src/core/SealedBidAuctionApiServer.ts
@@ -1,0 +1,49 @@
+import express from "express";
+import cors from "cors";
+import {Server} from "http";
+
+import {AuctionController} from "../api/controller/AuctionController.ts";
+import {WinstonLogger} from "./utils/WinstonLogger.ts";
+
+export class SealedBidAuctionApiServer {
+
+    private logger: WinstonLogger = new WinstonLogger(SealedBidAuctionApiServer.name);
+
+    private apiServer: Server | null = null;
+
+    async start(port: number) {
+        if (this.apiServer) {
+            this.logger.warn("API server is already running");
+            return;
+        }
+
+        const app = express();
+        const router = express.Router();
+
+        // Middleware
+        app.use(cors());
+        app.use(express.json());
+
+        // Register routes
+        const auctionController = new AuctionController();
+
+        router.post("/auction",
+            (req, res) => auctionController.open(req, res)
+        );
+
+        app.use("/api", router);
+
+        this.apiServer = app.listen(port, () => {
+            this.logger.info(`API server started on port ${port}`);
+        });
+    }
+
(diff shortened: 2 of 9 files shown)


tee-aps/sealed-bid-auction/src/api/controller/AuctionController.ts
```diff
@@ -0,0 +1,8 @@
+import type {Request, Response} from "express";
```
aharvet, line 1: When I mentioned using Bun for it I meant using the built in http server for performance.
Not a big deal though, express is not the fastest but for poc it will be okay.
But I have a feeling that we will keep it for longer than this, if we want to secure headroom for performance (in case we won't have time for rewrite in rust) the time is now to switch as it's super trivial still.

tee-aps/sealed-bid-auction/src/core/SealedBidAuctionApiServer.ts
```diff
@@ -0,0 +1,49 @@
...
+        const router = express.Router();
+
+        // Middleware
+        app.use(cors());
+        app.use(express.json());
+
+        // Register routes
+        const auctionController = new AuctionController();
+
+        router.post("/auction",
```
aharvet, line 30: This is for frontend and aggregator right ? This should be a get.

tee-aps/sealed-bid-auction/src/core/SealedBidAuctionApiServer.ts
```diff
@@ -0,0 +1,49 @@
...
+            (req, res) => auctionController.open(req, res)
+        );
+
+        app.use("/api", router);
+
+        this.apiServer = app.listen(port, () => {
+            this.logger.info(`API server started on port ${port}`);
+        });
+    }
+
```
aharvet, line 40: We also need a websocket server for price streaming with solvers. I remember you mentioned concerns about security for it, we can use wss which is the equivalent of https.

aharvet (commented):
Good base. I added a few comments, we can discuss around those :)

Also I am not a big fan of monorepos, I feel like it would be better to have a separate repo for this backend and not cross the PRs between multiple softwares. But I guess there are reasons to go for a monorepo.

tee-aps/sealed-bid-auction/src/core/SealedBidAuctionApiServer.ts
kss-t1: Yeah right now it's a dummy method anyway. Will change as I implement it in another PR.

tee-aps/sealed-bid-auction/src/core/SealedBidAuctionApiServer.ts
kss-t1: Sure, this will be also done in other PR.

tee-aps/sealed-bid-auction/src/core/SealedBidAuctionApiServer.ts
aharvet: Makes sense !

tee-aps/sealed-bid-auction/src/api/controller/AuctionController.ts
aharvet: Just to precise my thoughts: I was planning on using ElysiaJs that makes the syntax very similar to express. Also simplifies syntax for websocket.

PR #140, opened 2025-07-22
Feat/enforce auction winner

aharvet:
## Description

This PR add a check for the data in fillInstruction field to match with account doing the fill.
Used to ensure only the auction winner can fill the intent.
The contracts can still be used as usual, without specifying anything in that field and let the fill opened.

## Motivation and Context

We need a way to enforce the result of offchain auction.

## How Has This Been Tested?

Add a test suite specifically for this feature.

## Checklist:

### DIFF
--- a/contracts/src/7683/BasicSwap7683.sol
+++ b/contracts/src/7683/BasicSwap7683.sol
@@ -71,6 +71,7 @@ abstract contract BasicSwap7683 is Base7683 {
     error InvalidOrderDomain();
     error InvalidDomain();
     error InvalidSender();
+    error InvalidSolver(address enforcedSolver);
 
     // ============ Modifiers ============
 
@@ -121,7 +122,7 @@ abstract contract BasicSwap7683 is Base7683 {
         uint32 _messageOrigin,
         bytes32 _messageSender,
         bytes32 _orderId,
-        address _receiver
+        bytes32 _receiver
     )
         internal
         virtual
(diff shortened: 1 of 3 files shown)


contracts/src/7683/BasicSwap7683.sol
```diff
@@ -372,6 +373,11 @@ abstract contract BasicSwap7683 is Base7683 {
         if (_orderId != OrderEncoder.id(orderData)) revert InvalidOrderId();
         if (block.timestamp > orderData.fillDeadline) revert OrderFillExpired();
         if (orderData.destinationDomain != _localDomain()) revert InvalidOrderDomain();
+        // Enforce auction winner
```
evchip, line 376: My comment is somewhat outside the scope of this PR, but:
the `fill` method has a `_fillerData` param that is passed to this method `_fillOrder` but not used. We are however setting it on `filledOrders`, which is the param we use for settlements via xChainRead, i.e. whatever address the solver passes into `_fillerData` will be the address that receives the user's escrow on the source chain. I think this is a good feature since it lets solvers set a different address to receive funds, but we should be more explicit about it by:
1. removing the param from `_fillOrder`
2. changing the name to something like `depositRecipient`

no need to address in this PR though

evchip (approved):
LGTM, thanks

contracts/src/7683/BasicSwap7683.sol
aharvet: 1. Yes lot's of grooming to do it those contracts !
2. fillerData name is actually from the erc, and it's meant to be arbitrary data (for eg. that's where we would put the sig for winner enforcement with auction after intent open) so I think we should rather describe the format and fields expected in those bytes in comments

PR #144, opened 2025-07-25
Feat/sealed bid auction/price stream

kss-t1:
## Description

Implements some key services of Sealed Bid App. More specifically:
- Solver PriceBook which allows Solvers to stream their prices using WebSocket
- Auction feature which lets user retrieve the best price and a REST method to expose that information to the frontend
- Integration test for basic WebSocket logic

## Motivation and Context

It is a part of Sealed Bid App which provides a non-EVM way to establish the best price for filling 7683 Intent among competing Intent Solvers.

## How Has This Been Tested?

Tested on local using integration test.

-   [ ] Docs

## Checklist:

-   [ ] If my change requires additions or updates to any deployment scripts to ensure that the protocol is functional (Makefile, Dockerfile, Forge Script, etc.), I have made these changes, and in either case, have checked this box to attest to my assessment of this requirement with regard to my change.


tee-apps/sealed-bid-auction/src/core/AuctionService.ts
```diff
@@ -0,0 +1,69 @@
...
+    private findPricesForAskedTokens(request: AuctionRequest) {
+        return this.solverPricebook.getCurrentPrices().flatMap(
+            priceItems => priceItems.filter(
+                priceItem =>
+                    priceItem.srcTokenAddresses.includes(request.srcTokenAddress) &&
+                    priceItem.dstTokenAddress.includes(request.dstTokenAddress) &&
+                    this.getCorrectInterval(priceItem, request.amountIn) !== undefined
+            ).map(priceItem => {
+                return {
+                    price: this.getCorrectInterval(priceItem, request.amountIn)!.price,
```
aharvet: This should not return the price for highest interval but the overall amountOut.
Meaning, computing the amountOut for each intervals until highest one added up.
Example:
0 -> 99 : price = 1.5
100 -> 299 : price = 2
amountIn = 210
Then we should return amountOut = 99 * 1.5 + (210 - 99) * 2

aharvet (changes requested):
Everything good appart from returned data in quote.

tee-apps/sealed-bid-auction/src/core/AuctionService.ts
```diff
@@ -0,0 +1,86 @@
...
+        let currentInterval = 0;
+        let amountOut = 0n;
+        const finalIndex = this.getFinalIntervalIndex(priceItem, amountIn)!;
+
+        do {
+            const currInterval = priceItem.intervals[currentInterval]!;
+            if (currInterval !== priceItem.intervals[finalIndex]!) {
+                amountOut += currInterval.price * (currInterval.range.max - currInterval.range.min);
+            } else {
+                amountOut += currInterval.price * (amountIn - currInterval.range.min - (finalIndex === 0 ? 0n : 1n));
```
aharvet, line 80: Out of curiosity, why did you have to substract 1 here ?

tee-apps/sealed-bid-auction/src/core/AuctionService.ts
```diff
@@ -0,0 +1,86 @@
...
+        return this.solverPricebook.getCurrentPrices().flatMap(
+            priceItems => priceItems.filter(
+                priceItem =>
+                    priceItem.srcTokenAddresses.includes(request.srcTokenAddress) &&
+                    priceItem.dstTokenAddress.includes(request.dstTokenAddress) &&
+                    this.getFinalIntervalIndex(priceItem, request.amountIn) !== undefined
+            ).map(priceItem => {
+                return {
+                    amountOut: this.calculateAmountOut(priceItem, request.amountIn),
+                    solverAddress: priceItem.solverAddress
```
aharvet, line 64: I would like to be extra specific in the naming of this field, to make sure solvers make no mistake, and call this `settlementReceiverAddress` as solver can use a different address on source chain and destination chain.

aharvet (approved):

tee-apps/sealed-bid-auction/src/core/AuctionService.ts
kss-t1: Ok I renamed it in all the types for consistency.

tee-apps/sealed-bid-auction/src/core/SolverPriceBook.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export class SolverPriceBook {
+    private prices: Map<string, PriceBookEntry> = new Map<string, PriceBookEntry>();
+
+    constructor(private readonly priceListTTL: number = 600) {}
+
+    public getCurrentPrices(): ImmutableList<PriceListItem[]> {
+        return ImmutableList(
+            this.prices.entries()
+                .filter(([_key, value]) => value.timestamp > Date.now() + this.priceListTTL)
```
w-woloszyn, line 18: Shouldn't that be `.filter(([_key, value]) => value.timestamp + this.priceListTTL > Date.now())`? Since timestamp is recorded with `Date.now()` at the moment an update occurs, adding `priceListTTL` to the current time will make the right-hand side always ahead of the timestamp. To correctly exclude expired entries, shouldn't it compare timestamp + TTL against Date.now()?

kss-t1: You are right. Will fix in open PR

tee-apps/sealed-bid-auction/src/core/AuctionService.ts
```diff
@@ -0,0 +1,86 @@
...
+            };
+        }
+    }
+
+    private getFinalIntervalIndex(priceListItem: PriceListItem, amount: bigint): number | undefined {
+        return priceListItem.intervals.findIndex(interval => interval.range.min <= amount && interval.range.max >= amount);
+    }
+
+    private chooseBestPrice(pricesForAskedTokens: Immutable.List<Price>): Price | null {
+        let bestPrice = null;
```
w-woloszyn, line 38: `bestPrice` is redeclared inside the if block, shadowing the variable defined at the top of the function

tee-apps/sealed-bid-auction/src/core/AuctionService.ts
```diff
@@ -0,0 +1,86 @@
...
+    private calculateAmountOut(priceItem: PriceListItem, amountIn: bigint): bigint {
+        let currentInterval = 0;
+        let amountOut = 0n;
+        const finalIndex = this.getFinalIntervalIndex(priceItem, amountIn)!;
+
+        do {
+            const currInterval = priceItem.intervals[currentInterval]!;
+            if (currInterval !== priceItem.intervals[finalIndex]!) {
+                amountOut += currInterval.price * (currInterval.range.max - currInterval.range.min);
+            } else {
```
w-woloszyn, line 79: Looks like we’re short‑changing every full interval by one unit.

Why? We treat ranges as *inclusive* on both ends:

* `getFinalIntervalIndex` uses `<=` and `>=`, so an amount equal to `range.max` still belongs to that interval.
* In price‑list validation we insist that `prev.max + 1 === next.min`, which only makes sense if `prev.max` itself is part of the previous interval.
* There’s a little “`+1n`” hack later for the last interval that tries to patch the missing unit.

Given that, an interval `[min, max]` contains `max - min + 1` units, not `max - min`.

```suggestion
// inclusive range: add +1 to capture the max element
const size = currInterval.range.max - currInterval.range.min + 1n;
amountOut += currInterval.price * size;

File: tee-apps/sealed-bid-auction/src/core/AuctionService.ts [hunk 7]
```diff
@@ -0,0 +1,86 @@
...
+
+    private calculateAmountOut(priceItem: PriceListItem, amountIn: bigint): bigint {
+        let currentInterval = 0;
+        let amountOut = 0n;
+        const finalIndex = this.getFinalIntervalIndex(priceItem, amountIn)!;
+
+        do {
+            const currInterval = priceItem.intervals[currentInterval]!;
+            if (currInterval !== priceItem.intervals[finalIndex]!) {
+                amountOut += currInterval.price * (currInterval.range.max - currInterval.range.min);
```
w-woloszyn, line 78: Looks like we’re short‑changing every full interval by one unit.

Why? We treat ranges as *inclusive* on both ends:

* `getFinalIntervalIndex` uses `<=` and `>=`, so an amount equal to `range.max` still belongs to that interval.
* In price‑list validation we insist that `prev.max + 1 === next.min`, which only makes sense if `prev.max` itself is part of the previous interval.
* There’s a little “`-1n`” hack later for the last interval that tries to patch the missing unit.

Given that, an interval `[min, max]` contains `max - min + 1` units, not `max - min`.

```suggestion
// inclusive range: add +1 to capture the max element
const size = currInterval.range.max - currInterval.range.min + 1n;
amountOut += currInterval.price * size;

File: tee-apps/sealed-bid-auction/src/core/AuctionService.ts (hunk 2 above)
Comment (w-woloszyn) [inline tee-apps/sealed-bid-auction/src/core/AuctionService.ts:80]: This `- 1n` adjustment is trying to make up for the earlier off‑by‑one,
but once we count the full intervals correctly we don’t need any patch
at all.  For an inclusive range the slice from `range.min` up to
`amountIn` (inclusive) has `amountIn - range.min + 1` units.

```suggestion
// inclusive slice: count units from min up to amountIn
const size = amountIn - currInterval.range.min + 1n;
amountOut += currInterval.price * size;

File: tee-apps/sealed-bid-auction/src/core/AuctionService.ts [hunk 8]
```diff
@@ -0,0 +1,86 @@
...
+            ).map(priceItem => {
+                return {
+                    amountOut: this.calculateAmountOut(priceItem, request.amountIn),
+                    solverAddress: priceItem.solverAddress
+                };
+            })
+        )
+    }
+
+    private calculateAmountOut(priceItem: PriceListItem, amountIn: bigint): bigint {
```
w-woloszyn, line 70: Look at my comments bellow that suggest a fix of a potential off‑by‑one that shows up whenever an order
spans two or more price intervals.  Our ranges are inclusive
(`<=` / `>=` and `prev.max + 1 == next.min` in validation), so:

* Full interval size = `max - min + 1`.
* Final partial interval size = `amountIn - min + 1`.

Current code uses `max - min` for full intervals **and** subtracts
`1n` for the last slice, so `amountOut` is always too small.

After applying the two suggestions please add tests:

* amountIn at an interval boundary
* amountIn that spans 2+ intervals
* first‑interval‑only case

Those should fail before this patch and pass afterwards.

tee-apps/sealed-bid-auction/src/core/AuctionService.ts
kss-t1: This is out of date.

tee-apps/sealed-bid-auction/src/core/AuctionService.ts
kss-t1: This is out of date - was already spotted and fixed.

tee-apps/sealed-bid-auction/src/core/AuctionService.ts
kss-t1: Out of date.

PR #146, opened 2025-07-28
Feat/enforce auction winner v2

aharvet:
## Description

Add winner bid enforcement with a commit transaction sent from backend.

## Motivation and Context

We need to way to force solvers to respect the bid the commit to, after the open intent transaction has been done.

## How Has This Been Tested?

Test scripts.

-   [ ] Docs

## Checklist:

### DIFF
--- a/contracts/src/test/7683/T1XChainReaderBaseTestSetup.sol
+++ b/contracts/src/test/7683/T1XChainReaderBaseTestSetup.sol
@@ -64,6 +64,7 @@ contract T1XChainReaderBaseTestSetup is BaseTest {
             destinationDomain: destination,
             destinationSettler: address(l2T1ERC7683).addressToBytes32(),
             fillDeadline: uint32(block.timestamp + 100),
+            closedAuction: false,
             data: new bytes(0)
         });
     }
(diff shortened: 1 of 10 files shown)


contracts/src/test/7683/T1XChainReaderBaseTestSetup.sol
```diff
@@ -64,6 +64,7 @@ contract T1XChainReaderBaseTestSetup is BaseTest {
             destinationDomain: destination,
             destinationSettler: address(l2T1ERC7683).addressToBytes32(),
             fillDeadline: uint32(block.timestamp + 100),
+            closedAuction: false,
```
evchip, line 67: What is the significance of this flag? Would be good to have some documentation on it

aharvet: It will be true if we run a closed auction offchain on our auction backend and false if the intent is free to be filled by anyone. But maybe there could be a better name ?

evchip: `isExclusiveFill`? idk

it doesn't look like we have any test where `closedAuction` is set to true. Is this on purpose?

aharvet: Yes why not, I am going to think about it.

We do have such tests in this file https://github.com/t1protocol/t1/pull/146/files#diff-6a4d69cbb09ccb6052ee833d72a67836a6bf44354a96c39858396427bc25d118R133

evchip (approved):

PR #148, opened 2025-07-31
Chore/contracts grooming

aharvet:
## Description

Cleaning of the 7683 contracts in order to prepare audit.
- solidity version bump
- cleaning of old inheritance architecture
- grouping of functions
- split of permit2 logic
- add proper interface
- small improvements in namings
- improve comments
- remove dead code

## Motivation and Context

As audit is coming, we need to provide a decently readable code to facilitate and reduce auditors work.
Also as the code will be deployed on mainnet, we need it to be as futur proof as possible.

## How Has This Been Tested?

Standard testing suite is still passing.

-   [x] Style (style only changes)
-   [x] Docs

## Checklist:


evchip:
are we going to use Permit2?

contracts/src/7683/T1Permit2.sol
```diff
@@ -0,0 +1,128 @@
...
+import { IPermit2, ISignatureTransfer } from "@uniswap/permit2/src/interfaces/IPermit2.sol";
+import { ResolvedCrossChainOrder } from "../interfaces/IERC7683.sol";
+
+/// @title T1Permit2
+/// @notice Abstract contract that handles Permit2 integration for gasless cross-chain order transactions.
+/// @author BootNode
+/// @dev Provides functionality for signature-based token transfers using Uniswap's Permit2 system,
+/// enabling gasless order operations where users can sign permit messages instead of making on-chain approvals.
+/// This contract manages nonce tracking and signature validation for gasless transactions.
+abstract contract T1Permit2 {
```
evchip, line 14: are we going to use this?

contracts/src/7683/T1ERC7683.sol
```diff
@@ -170,20 +105,191 @@ contract T1ERC7683 is Base7683, OwnableUpgradeable, PausableUpgradeable {
...
+    )
+        internal
+        view
+        returns (ResolvedCrossChainOrder memory resolvedOrder, bytes32 orderId, uint256 nonce)
+    {
+        if (_orderType != OrderEncoder.orderDataType()) revert InvalidOrderType(_orderType);
+
+        // IDEA: _orderData should not be directly typed as OrderData, it should contain information that is not
+        // present on the type used for open the order. So _fillDeadline and _user should be passed as arguments
+        OrderData memory orderData = OrderEncoder.decode(_orderData);
```
evchip, line 206: i'm not following this. could you please explain?

contracts/src/7683/T1ERC7683.sol
```diff
@@ -170,20 +105,191 @@ contract T1ERC7683 is Base7683, OwnableUpgradeable, PausableUpgradeable {
...
+
+        if (orderData.originDomain != localDomain) revert InvalidOriginDomain(orderData.originDomain);
+
+        // bytes32 destinationSettler = _mustHaveRemoteCounterpart(orderData.destinationDomain);
```
evchip: If unneeded should we remove it? IIRC this was removed by me because we didn't want to constrain the design to two chains only, i.e. need to deploy a new 7683 on chain A when adding a new 7683 on chain B

contracts/src/7683/T1ERC7683.sol
```diff
@@ -1,120 +1,59 @@
...
+    mapping(bytes32 orderId => FilledOrder filledOrder) public filledOrders;
+    /// @notice Tracks the status of each order by its ID.
+    mapping(bytes32 orderId => Status status) public orderStatus;
     mapping(bytes32 requestId => uint256 batchIndex) public requestIdToBatchIndex;
-    /// @notice Maps request IDs to order IDs for cross-chain read requests for settlements
-    mapping(bytes32 => bytes32) public settlementReadRequestToOrderId;
     /// @notice Maps request IDs to order IDs for cross-chain read requests for refunds
     mapping(bytes32 => bytes32) public refundReadRequestToOrderId;
-    /// @notice Maps order IDs to verification status
-    mapping(bytes32 => bool) public orderVerified;
```
evchip, line 49: why is this no longer needed?

contracts/src/7683/T1Permit2.sol
aharvet: Can does advocate for gasless intent opening. But yes, permit2 is an outdated way to do it with 7702 now. Also it requires to have users approve permit2 contract prior to that which is not very common.
If you think we should get rid of this, I will happily proceed.

evchip: I'm fine to keep it if we actually implement it in 7683 (right now our scripts don't don't deploy nor set Permit2 on the 7683 contracts). Otherwise we should remove it. Either way is fine by me.

contracts/src/7683/T1ERC7683.sol
aharvet: Yes it is horribly worded, not sure why I was so unclear.
What I meant is that there are values in OrderData that are already present in the global Order struct like `user` and `sender`, `fillDeadline`, `originChainId` and `originDomain`.
So we don't need to have them in the OrderData struct and should remove them. But I left them for now as our offchain infra might use them.
I will improve the comment.

contracts/src/7683/T1ERC7683.sol
aharvet: Yes

contracts/src/7683/T1ERC7683.sol
aharvet: It was not used anywhere in the contract, and redundant with order status from offchain.

contracts/src/7683/T1ERC7683.sol
evchip: Got it, thanks

evchip (approved):

contracts/src/7683/T1Permit2.sol
aharvet: After thinking about it, permit2 is actually needed to comply with erc 7683 for the implementation of openFor, so probably we should keep it eventhough we don't use it.
We can still add permit2 address to deployment scripts.

PR #155, opened 2025-08-06
Feature: xYield ERC-4626

evchip:
## Description
- create xYield contract, inheriting ERC-4626
  - add `deposit` and `depositFrom` methods for same chain and remote chain deposits, respectively
  - add method `updateTotals` with permission check for guardian (offchain program running in TEE) in order to update global shares and user positions
  - add placeholders for unimplemented method `rebalance`
- create scripts to deploy xYield contracts to Arbitrum and Base mainnets
- add deployment metadata for xYield v0.0.1 deployment

## Motivation and Context
Checkpoint with boilerplate functionality in order to deploy and unblock offchain work streams.

## How Has This Been Tested?
Fork tests and unit tests for `deposit` and `depositFrom` methods

## Checklist:


contracts/src/xYield/xYieldTest.t.sol
```diff
@@ -0,0 +1,172 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
```
aharvet: Maybe we should put those test files in the test folder ?

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,288 @@
...
+    uint256 public virtualTotalAssets;
+    uint256 public virtualTotalSupply;
+    bool public isActiveChain;
+
+    mapping(uint256 => address) public siblingVaults;
+
+    uint256 public minRebalanceGap;
+    uint256 public lastRebalanceTime;
+
+    event DepositRemote(address indexed sender, address indexed owner, uint256 assets, uint256 shares);
```
aharvet: Let's group this event with the other ones ?

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,288 @@
...
+    {
+        guardian = _guardian;
+        yieldProtocol = IERC4626(_yieldProtocol);
+        minRebalanceGap = 1 hours;
+
+        if (_yieldProtocol != address(0)) {
+            underlying.approve(_yieldProtocol, type(uint256).max);
+        }
+
+        _transferOwnership(msg.sender);
```
aharvet: I think ownership is already transfered to sender in Ownable isn't it ?

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    function deposit(
+        uint256 assets,
```
aharvet: assets is not very descriptive to me here, I would rather go for amount

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,288 @@
...
+        whenNotPaused
+        returns (uint256)
+    {
+        if (assets == 0) revert ZeroAmount();
+        if (!isActiveChain) revert DepositOnInactiveChain();
+
+        uint256 shares = previewDeposit(assets);
+        _deposit(msg.sender, receiver, assets, shares);
+
+        return shares;
```
aharvet, line 71: How could this return value be used ?

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    function deposit(
+        uint256 assets,
+        address receiver
+    )
+        public
+        virtual
+        override
+        nonReentrant
```
aharvet: Do we actually need to use nonReentrant ?
There is no ether transfer which is the biggest reentrancy risk factor, the interaction with external contracts are at the very end which respects check effect interaction pattern and those are white listed.

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,288 @@
...
+        yieldProtocol.deposit(assets, address(this));
+
+        emit Deposit(caller, receiver, assets, shares);
+    }
+
+    // NOTE - for remote deposits we update virtualTotalAssets after deposit to prevent share price inflation before
+    // global shares is updated
+    function _depositFrom(address caller, address receiver, uint256 assets, uint256 shares) internal {
+        IERC20(asset()).transferFrom(caller, address(this), assets);
+        yieldProtocol.deposit(assets, address(this));
```
aharvet: We don't account for total assets and supply here because it's already done on initial deposit chain right ?

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,288 @@
...
+        if (!isActiveChain) revert DepositOnInactiveChain();
+
+        uint256 shares = previewDeposit(assets);
+        _deposit(msg.sender, receiver, assets, shares);
+
+        return shares;
+    }
+
+    // called on behalf of a user who has deposited from a remote chain
+    function depositFrom(uint256 assets, address receiver) public nonReentrant whenNotPaused returns (uint256 shares) {
```
aharvet: Shouldn't we restrein access to our infra only ? to avoid anyone to mess up the crosschain accounting
Or maybe use something to prove an previous deposit has been done on another chain ?

aharvet (commented):
NOT FINISHED

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,288 @@
...
+        address[] calldata _recipients,
+        uint256[] calldata _amounts
+    )
+        external
+        onlyGuardian
+    {
+        // TODO - update incrementally like virtualTotalAssets using proof of remote supply change
+        virtualTotalSupply = totalSupply_;
+        _updateBalances(_recipients, _amounts);
+        virtualTotalAssets = this.totalAssets();
```
aharvet: Is there a reason why we need to use this. here and below instead of accessing the variable directly ?

contracts/src/xYield/xYieldTest.t.sol
evchip: fixed, thanks.

contracts/src/xYield/xYieldVault.sol
evchip: fixed, thanks.

contracts/src/xYield/xYieldVault.sol
evchip: fixed, thanks.

contracts/src/xYield/xYieldVault.sol
evchip: If another contract (e.g. a vault or smart account) calls deposit, they get back the number of shares minted to them from their deposit.

contracts/src/xYield/xYieldVault.sol
evchip: total assets and supply will be updated in subsequent txs `updateTotals` called by the TEE guardian. The `virtualTotalAssets` and `virtualTotalSupply` will be updated with the same values across all chains. The user who deposited from the remote chain will be minted his share tokens there. This prevents a scenario where users who already have shares in the pool are able to withdraw the remote deposit user's assets.

contracts/src/xYield/xYieldVault.sol
evchip: This will be called by solvers, I believe. The deposit is actually being done on this chain, it's simply an affordance to users who want their share tokens minted on another chain. I don't think we need to restrict it. After the solver deposits, we will pick up the event offchain and mint the shares for them on the remote chain.

But I see that I should add the chain ID as a param so we know which chain to mint to, and also to check whether it is a supported chain.

contracts/src/xYield/xYieldVault.sol
evchip: good catch, fixed

contracts/src/xYield/xYieldVault.sol
evchip: This is Ownable2Step. The first step is `_transferOwnership`, the second step is for the recipient to call `acceptOwnership`

contracts/src/xYield/xYieldVault.sol
aharvet: Ok makes sense

contracts/src/xYield/xYieldVault.sol
aharvet: I mean Ownable2Step inherits Ownable, and `_transferOwnership(_msgSender())` is already called in Ownable constructor

contracts/src/xYield/xYieldVault.sol
aharvet: Ok makes sense

contracts/src/xYield/xYieldVault.sol
aharvet: Ok makes sense, I think I'm just not 100% aware of the backend role yet

contracts/src/xYield/xYieldVault.sol
evchip: good point, fixed

contracts/src/xYield/xYieldVault.sol
evchip: Got it, removed

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,292 @@
...
+
+    mapping(uint64 => address) public siblingVaults;
+
+    uint256 public minRebalanceGap;
+    uint256 public lastRebalanceTime;
+
+    error NotGuardian();
+    error ZeroAmount();
+    error InvalidChain();
+    error RebalanceTooFrequent();
```
aharvet: Error not used. Maybe it's for future ?

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,292 @@
...
+import { Pausable } from "@openzeppelin/contracts/security/Pausable.sol";
+
+contract xYieldVault is ERC4626, Ownable2Step, ReentrancyGuard, Pausable {
+    using Math for uint256;
+
+    address public guardian;
+    IERC4626 public yieldProtocol;
+
+    uint256 public virtualTotalAssets;
+    uint256 public virtualTotalSupply;
```
aharvet, line 22: Not necessarily an AI but I have a hard time wrapping my head around those variable names after multiple read of the contract. Maybe `virtualTotalDeposit` and `virtualTotalLiquidity` could be more explicit ?

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,292 @@
...
+            _mint(_recipients[i], _amounts[i]);
+        }
+    }
+
+    function _deposit(address _caller, address _receiver, uint256 _amount, uint256 _shares) internal virtual override {
+        virtualTotalAssets += _amount;
+        virtualTotalSupply += _shares;
+        _mint(_receiver, _shares);
+
+        IERC20(asset()).transferFrom(_caller, address(this), _amount);
```
aharvet: We should use `safeERC20` for transfers.

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,292 @@
...
+        return shares;
+    }
+
+    // called on behalf of a user who has deposited from a remote chain
+    function depositFrom(
+        uint256 _amount,
+        address _receiver,
+        uint64 _chainId
+    )
+        public
```
aharvet: I think we can set this one as external ?

contracts/src/xYield/xYieldVault.sol
aharvet: @evchip just to make sure, so if a user would call this one directly (and not a solver) he would just receive the shares on a different chain right ?

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    function mint(
+        uint256 _shares,
+        address _receiver
+    )
+        public
+        virtual
+        override
+        nonReentrant
```
aharvet: I think my previous comment about nonReentrant applies to all of them

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,292 @@
...
+        }
+
+        // Should only be called on active chains where assets exist
+        if (!isActiveChain) revert WithdrawOnInactiveChain();
+
+        yieldProtocol.withdraw(_amount, address(this), address(this));
+        virtualTotalAssets -= _amount;
+        virtualTotalSupply -= _shares;
+        _burn(_owner, _shares);
+        IERC20(asset()).transfer(_receiver, _amount);
```
aharvet: We should use safeERC20 for transfers.

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,292 @@
...
+    function _depositFrom(
+        address _caller,
+        address _receiver,
+        uint256 _amount,
+        uint256 _shares,
+        uint64 _chainId
+    )
+        internal
+    {
+        IERC20(asset()).transferFrom(_caller, address(this), _amount);
```
aharvet: We should use safeERC20 for transfers.

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,292 @@
...
+import { ERC4626 } from "@openzeppelin/contracts/token/ERC20/extensions/ERC4626.sol";
+import { IERC4626 } from "@openzeppelin/contracts/interfaces/IERC4626.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import { Math } from "@openzeppelin/contracts/utils/math/Math.sol";
+import { Ownable2Step } from "@openzeppelin/contracts/access/Ownable2Step.sol";
+import { ReentrancyGuard } from "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+import { Pausable } from "@openzeppelin/contracts/security/Pausable.sol";
+
+contract xYieldVault is ERC4626, Ownable2Step, ReentrancyGuard, Pausable {
```
aharvet, line 14: Should we have an interface for it ?

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    function updateVirtualTotalAssets() external onlyGuardian {
+        virtualTotalAssets = totalAssets();
+    }
+
+    // TODO - submit proofs for remote deposits
+    function _updateBalances(address[] calldata _recipients, uint256[] calldata _amounts) private {
+        if (_recipients.length != _amounts.length) revert LengthMismatch();
+        for (uint256 i = 0; i < _recipients.length; i++) {
```
aharvet, line 176: I am afraid this won't scale, f we get a lot of users this will be very expensive.
What do you think ?

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,292 @@
...
+    IERC4626 public yieldProtocol;
+
+    uint256 public virtualTotalAssets;
+    uint256 public virtualTotalSupply;
+    bool public isActiveChain;
+
+    mapping(uint64 => address) public siblingVaults;
+
+    uint256 public minRebalanceGap;
+    uint256 public lastRebalanceTime;
```
aharvet: Is it meant that we don't use these variables yet ?

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,292 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
```
aharvet: Version should be fixed

contracts/src/test/xYield/xYieldTest.t.sol
```diff
@@ -0,0 +1,186 @@
...
+        uint256 depositAmount = 100 * 10 ** 18;
+
+        vm.prank(user1);
+        uint256 shares = vault.deposit(depositAmount, user1);
+
+        assertEq(vault.balanceOf(user1), shares, "shares balance of user1");
+        assertEq(vault.totalSupply(), shares, "total supply");
+        assertEq(vault.totalAssets(), depositAmount, "total assets");
+        assertEq(usdc.balanceOf(user1), INITIAL_DEPOSIT - depositAmount, "usdc balance of user1");
+    }
```
aharvet, line 82: Maybe we should also test other functions like depositFrom, mint, redeem... at least for non regression ?

aharvet (commented):
A few more comments

contracts/src/xYield/xYieldVault.sol
```diff
@@ -0,0 +1,292 @@
...
+    using Math for uint256;
+
+    address public guardian;
+    IERC4626 public yieldProtocol;
+
+    uint256 public virtualTotalAssets;
+    uint256 public virtualTotalSupply;
+    bool public isActiveChain;
+
+    mapping(uint64 => address) public siblingVaults;
```
aharvet, line 25: Is it relevant to use uint64 here as there's no gas saving ?

evchip: most ethereum clients use uint64 for chainId. i'm just sticking to that convention

aharvet: Okay makes sense

contracts/src/xYield/xYieldVault.sol
evchip: Yes, exactly

contracts/src/xYield/xYieldVault.sol
evchip: it's a good point. there are two ways that we could get away with users minting share tokens:
1. mint them when users initiate a remote deposit request. if the intent fails, we burn it for them. intuitively this seems a bit risky because we (temporarily) have share tokens minted before the underlying asset has reached the vault. maybe it's fine as long as we have a way to make it atomic.
2. we post a merkle root to each chain which proves their deposit landed, and expose proofs that the user can submit which if valid allow them to mint share tokens.

let me know which approach you think is better, or if you have a better idea.

contracts/src/test/xYield/xYieldTest.t.sol
evchip: `depositFrom` is covered in the fork test. I can do mint and redeem, but not a high priority IMO. I'll make a TODO inline.

contracts/src/xYield/xYieldVault.sol
evchip: done

contracts/src/xYield/xYieldVault.sol
evchip: done

contracts/src/xYield/xYieldVault.sol
evchip: done

contracts/src/xYield/xYieldVault.sol
evchip: I'm just following the 4626 convention but prefixing to `virtual` to make clear it's not necessarily the same as `totalAssets` and `totalSupply`. Let me do a write up on this so it's more clear.

contracts/src/xYield/xYieldVault.sol
evchip: removed

contracts/src/xYield/xYieldVault.sol
evchip: removed

contracts/src/xYield/xYieldVault.sol
evchip: done

contracts/src/xYield/xYieldVault.sol
evchip: done

contracts/src/xYield/xYieldVault.sol
evchip: done

contracts/src/xYield/xYieldVault.sol
evchip: removed

contracts/src/xYield/xYieldVault.sol
aharvet: I'm fine with virtual, it's rather the assets and supply terms. But no worries I just wanted to mention it in case you feel the same way.

contracts/src/xYield/xYieldVault.sol
aharvet: I thought of 2 too. I agree that 1 is risky.
I just don't know if introducing this extra step for users resonate with the whole point of the product.

contracts/src/test/xYield/xYieldTest.t.sol
aharvet: Sure

aharvet (approved):
LGTM. We just need to keep in mind the loop in _updateBalances topic for next iterations.

PR #157, opened 2025-08-07
Feat/sealed bid auction/cors support

kss-t1:
## Description
CORS support for a method that frontend will use.
Also fixes some bugs:
- incorrect Time To Leave for `PriceList`s
- incorrect parsing of the BigInt amountIn
- incorrect assumption that amountIn should be an amount without decimals

## How Has This Been Tested?

on v07 and with automatic tests

-   [ ] Docs

## Checklist:


tee-apps/sealed-bid-auction/src/api/AuctionController.ts
```diff
@@ -14,13 +20,13 @@ export class AuctionController {
             const auctionQuote = this.auctionService.preauction(auctionRequest);
 
             if (auctionQuote) {
-                return new Response(serialize(auctionQuote), {status: 200});
+                return new Response(serialize(auctionQuote), {status: 200, headers: CORS_HEADERS});
```
w-woloszyn, line 23: HTTP 204 must not include a response body

tee-apps/sealed-bid-auction/src/api/AuctionApiServer.ts
```diff
@@ -38,7 +38,10 @@ export class AuctionApiServer {
             } : {},
             routes: {
                 "/healthcheck": new Response("OK"),
-                "/api/preauction": req => this.auctionController.preauction(req),
+                "/api/preauction": {
+                    GET: _req => new Response("Use POST", {status: 405}),
```
w-woloszyn: The 405 response should include CORS headers, otherwise the browser might block the error. Can include include `Allow` to show supported methods to users

```suggestion
                    GET: _req => new Response("Use POST", {
                      status: 405,
                      headers: { ...CORS_HEADERS, "Allow": "POST, OPTIONS" },
})
```

w-woloszyn: On top of that, browsers preflight `POST` with JSON/custom headers. Please add an `OPTIONS` handler so the request isn’t blocked by CORS. Also make sure these responses carry CORS headers.

tee-apps/sealed-bid-auction/src/api/AuctionApiServer.ts
```diff
@@ -38,7 +38,10 @@ export class AuctionApiServer {
             } : {},
             routes: {
                 "/healthcheck": new Response("OK"),
-                "/api/preauction": req => this.auctionController.preauction(req),
+                "/api/`preauction": {
```
princet1protocol: what is the purpose of this backtick?

tee-apps/sealed-bid-auction/src/api/AuctionController.ts
```diff
@@ -14,13 +20,17 @@ export class AuctionController {
...
             } else {
-                return new Response("No quote found for this pair", {status: 204});
+                return new Response(null, {status: 204, headers: CORS_HEADERS});
```
princet1protocol, line 25: why do we remove the response message?

tee-apps/sealed-bid-auction/src/api/AuctionApiServer.ts
kss-t1: It was there by mistake, already fixed

tee-apps/sealed-bid-auction/src/api/AuctionController.ts
kss-t1: because as per [HTTP specification](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/204), 204 Response should have no body. it seems to me that sending `204` gives client information about what happened - request was processed successfully and produced no results (no quote) .
if you don't think it is appropiate, let me know and we can change it to whatever you suggest

princet1protocol: i think i would prefer a message so that the frontend can react correspondingly, we may have to change away from 204.
for now its fine

princet1protocol (dismissed):

princet1protocol (approved):

PR #161, opened 2025-08-15
Feat/enforce auction winner v3

aharvet:
## Description

Implement auction winner enforcement with signature on destination chain.

## Motivation and Context

We concluded it's the best approach.

## How Has This Been Tested?

Updated closed auction test suite.

-   [ ] Docs

## Checklist:

### DIFF
--- a/contracts/src/7683/T1ERC7683.sol
+++ b/contracts/src/7683/T1ERC7683.sol
@@ -39,8 +39,6 @@ contract T1ERC7683 is IT1ERC7683, T1Permit2, OwnableUpgradeable, PausableUpgrade
     mapping(bytes32 requestId => uint256 batchIndex) public requestIdToBatchIndex;
     /// @notice Maps request IDs to order IDs for cross-chain read requests for refunds
     mapping(bytes32 => bytes32) public refundReadRequestToOrderId;
-    /// @notice Maps order IDs to their winning bid information
-    mapping(bytes32 orderId => Bid winnerBid) public orderToBid;
     /// @notice Maps request IDs to order IDs for cross-chain read requests for settlements
     mapping(bytes32 => bytes32) public settlementReadRequestToOrderId;
     address public counterpart;
@@ -364,20 +362,7 @@ contract T1ERC7683 is IT1ERC7683, T1Permit2, OwnableUpgradeable, PausableUpgrade
 
             for (uint256 i = 0; i < _orderIds.length; i++) {
                 if (_settled) {
-                    (uint256 amountOut, address settlementReceiver) =
-                        abi.decode(_ordersFillerData[i], (uint256, address));
-                    Bid memory winnerBid = orderToBid[orderId];
-
-                    if (
-                        orderData.closedAuction == true
-                            && (settlementReceiver != winnerBid.settlementReceiver || amountOut != winnerBid.amountOut)
-                    ) {
-                        revert InvalidFill(
-                            settlementReceiver, winnerBid.settlementReceiver, amountOut, winnerBid.amountOut
-                        );
-                    }
-                    delete orderToBid[orderId];
-
+                    (, address settlementReceiver) = abi.decode(_ordersFillerData[i], (uint256, address));
                     _handleSettleOrder(
                         orderData.destinationDomain, orderData.destinationSettler, _orderIds[i], settlementReceiver
                     );
@@ -550,15 +535,6 @@ contract T1ERC7683 is IT1ERC7683, T1Permit2, OwnableUpgradeable, PausableUpgrade
 
             emit Refunded(orderId, orderSender);
         }
-    }
-
-    /// @notice Commits a winning bid for a specific order
-    /// @dev Only accessible by owner
-    /// @param orderId The ID of the order to set the winner bid for
-    /// @param winnerBid The winning bid containing settlement receiver and amount out
-    function commitWinnerBid(bytes32 orderId, Bid calldata winnerBid) external onlyOwner {
-        orderToBid[orderId] = winnerBid;
-        emit WinnerBidCommited(orderId, winnerBid.settlementReceiver);
     }
 
     function pause() external onlyOwner {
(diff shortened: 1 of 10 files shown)


contracts/src/7683/T1ERC7683.sol
```diff
@@ -39,19 +41,32 @@ contract T1ERC7683 is IT1ERC7683, T1Permit2, OwnableUpgradeable, PausableUpgrade
...
     address public counterpart;
 
+    /// @notice EIP-712 typehash for fill authorization
+    bytes32 public constant FILL_AUTHORIZATION_TYPEHASH =
+        keccak256("FillAuthorization(bytes32 orderId,address filler,uint256 amountOut)");
+
     /// @notice Initializes the contract with the specified dependencies
     /// @param _permit2 The address of the permit2 contract
     /// @param _xChainRead The address of the cross-chain read contract
     /// @param _localDomain The local domain (chain id)
```
evchip, line 56: natspec please :)

contracts/src/7683/T1ERC7683.sol
```diff
@@ -39,19 +41,32 @@ contract T1ERC7683 is IT1ERC7683, T1Permit2, OwnableUpgradeable, PausableUpgrade
...
-    constructor(address _permit2, address _xChainRead, uint32 _localDomain) T1Permit2(_permit2) {
+    constructor(
+        address _permit2,
+        address _xChainRead,
+        uint32 _localDomain,
+        address _auctionBackend
+    )
```
evchip, line 63: `auctionProver`? `auctionWitness`?

contracts/src/7683/T1ERC7683.sol
```diff
@@ -258,32 +273,75 @@ contract T1ERC7683 is IT1ERC7683, T1Permit2, OwnableUpgradeable, PausableUpgrade
...
+    }
+
+    /// @dev Decodes filler data to extract amount and authorization
+    function _decodeFillerData(bytes calldata fillerData)
+        private
+        pure
+        returns (uint256 amountOut, bytes memory authorization)
+    {
+        if (fillerData.length > 64) {
+            (amountOut,, authorization) = abi.decode(fillerData, (uint256, address, bytes));
```
evchip, line 314: If we don't use `settlementReceiver` should we remove it from `fillerData`?

contracts/src/7683/T1ERC7683.sol
```diff
@@ -258,32 +273,75 @@ contract T1ERC7683 is IT1ERC7683, T1Permit2, OwnableUpgradeable, PausableUpgrade
...
+        bytes memory authorization
+    )
+        private
+        view
+    {
         if (orderId != OrderEncoder.id(orderData)) revert InvalidOrderId();
         if (block.timestamp > orderData.fillDeadline) revert OrderFillExpired();
         if (orderData.destinationDomain != localDomain) revert InvalidOrderDomain();
         if (amountOut < orderData.minAmountOut) revert AmountOutTooLow();
+        if (orderData.closedAuction && authorization.length > 0) {
```
evchip: What happens if it's a closed auction but the solver doesn't pass an authorization? Seems like it would pass the validation. Shouldn't it instead be:
```
if (orderData.closedAuction) {
 if (authorization.length > 0) {
  _verifyAuthorization(orderId, amountOut, authorization);
 } else {
  revert InvalidAuthorization();
 }
}
```

evchip (changes requested):

contracts/src/7683/T1ERC7683.sol
aharvet: You're right I just removed the `&& authorization.length > 0` and added 2 tests for this case

evchip: @aharvet I don't think you've pushed the change

aharvet: Yes I'll push the full changes for your review

contracts/src/7683/T1ERC7683.sol
aharvet: I like auction witness

contracts/src/7683/T1ERC7683.sol
aharvet: We do use it in the next parts of the logic, for settlement

contracts/src/7683/T1ERC7683.sol
```diff
@@ -39,19 +41,33 @@ contract T1ERC7683 is IT1ERC7683, T1Permit2, OwnableUpgradeable, PausableUpgrade
     mapping(bytes32 requestId => uint256 batchIndex) public requestIdToBatchIndex;
     /// @notice Maps request IDs to order IDs for cross-chain read requests for refunds
     mapping(bytes32 => bytes32) public refundReadRequestToOrderId;
-    /// @notice Maps order IDs to their winning bid information
-    mapping(bytes32 orderId => Bid winnerBid) public orderToBid;
     /// @notice Maps request IDs to order IDs for cross-chain read requests for settlements
     mapping(bytes32 => bytes32) public settlementReadRequestToOrderId;
+    address public immutable auctionWitness;
```
evchip, line 46: If I recall correctly, proxy storage cannot be immutable on ERC 1967 because it messes with the storage layout. I think this would cause trouble when you deploy.

evchip (dismissed):
Looks good, thanks. Please confirm whether using `immutable` will be an issue for ERC 1967 proxy deployment.

contracts/src/7683/T1ERC7683.sol
aharvet: I don't think so, immutable variables go to code so they are not in the storage layout. Do you mean for readability ?

Because ERC 1967 only imposes this for for logic, beacon and admin addresses storage slots, and this is handled in the proxy contract.

Or there is something else ?

evchip: makes sense, thanks

evchip (approved):

PR #164, opened 2025-08-20
feat: winner enforcement v3

kss-t1:
## Description
Implements Winner Enforcement v3 - so sending a signature to a Solver on a won auction.

## Motivation and Context
This is necessary to only allow auction-winning Solver to fill intent.

## How Has This Been Tested?
Local env and Base/Arbitrum Sepolia

-   [ ] Docs

## Checklist:

-   [ ] If my change requires a change to the documentation, I have updated the documentation accordingly, and in either case, have checked this box to attest to my assessment of this requirement with regard to my change.
-   [ ] If my change requires additional test coverage, I have created those tests accordingly, and in either case, have checked this box to attest to my assessment of this requirement with regard to my change.


tee-apps/sealed-bid-auction/src/api/types.ts
```diff
@@ -34,10 +32,11 @@ export interface AuctionQuote {
 }
 
 export interface AuctionResult {
+    type: string;
+    orderId: string;
     settlementReceiverAddress: string;
     amountOut: bigint;
-    orderId: string;
-    orderData: OrderData;
+    signature: string;
```
aharvet, line 39: Could it be relevant to already pass the value solver will have to pass in `fillerData` field already formatted ? --> abi.encode(uint256 amountOut, address settlementReceiver, bytes authorization) encoded as eip 712

Or do we consider that solvers will want to validates values first and build it themselves ?

tee-apps/sealed-bid-auction/src/blockchain/ViemIntentObserver.ts
```diff
@@ -87,21 +82,39 @@ export class ViemIntentObserver {
...
             if (winningPrice !== null && winningPrice.amountOut >= orderData.minAmountOut) {
+
                 const result: AuctionResult = {
+                    type: "winning-bid",
                     settlementReceiverAddress: winningPrice.settlementReceiverAddress,
                     amountOut: winningPrice.amountOut,
                     orderId,
-                    orderData
+                    signature: await this.signAuctionResult(
+                        BigInt(orderId), winningPrice.settlementReceiverAddress as `0x${string}`, winningPrice.amountOut
```
aharvet: Here the solver address is not the settlement receiver address anymore (which is source chain solver address) because of the switch from WE v2 to v3. It's now the address solver will use on destination chain. But I think it's just a naming change needed.

tee-apps/sealed-bid-auction/src/blockchain/ViemIntentObserver.ts
```diff
@@ -87,21 +82,39 @@ export class ViemIntentObserver {
...
+                this.logger.info(`I finished auction for orderId=[${orderId}] and notified solvers`);
                 break;
             }
             
             await new Promise((resolve) => setTimeout(resolve, this.auctionPollingInterval));
         }
     }
+
+    private async signAuctionResult(orderId: bigint, winningSolver: `0x${string}`, amountOut: bigint) : Promise<string> {
+        const encoded = encodeAbiParameters(
```
aharvet: Ok so my bad, I realise I didn't explicitly mention in the design that the values should also be encoded according to EIP 712 before being signed (sorry for being a smart contract nerd that assumes it's obvious).
I can take care of this if you want.

aharvet (changes requested):

tee-apps/sealed-bid-auction/src/blockchain/ViemIntentObserver.ts
kss-t1: it's the same address anyway... I would keep the name as is to avoid confusing Tokka - since they already stated integrating

aharvet: If you are sure they will use the same 100% yes that's okay. I just don't want this to be misleading at some point.

tee-apps/sealed-bid-auction/src/blockchain/ViemIntentObserver.ts
aharvet: @kss-t1 I pushed the commit

tee-apps/sealed-bid-auction/src/api/types.ts
kss-t1: As far as I understand, you implemented that, so resolving this conversation.

aharvet: No that's not what I did but it's ok, we can always add it if they request it

aharvet (approved):

PR #165, opened 2025-08-20
Feature/xyield erc 4626 withdraw

evchip:
### `xYieldVault.sol`
 - replace updateTotals params with struct
 - add withdrawFrom method for remote withdrawals
 - change updateBalances implementation to support burning or minting tokens
 - add fork test for remoteWithdraw method

## Motivation and Context
Facilitate same chain withdrawals. Remote withdrawals are partially implemented, but will need bridging capabilities to be fully functional. We are parking this for the time being.

## How Has This Been Tested?
Added fork and unit tests

## Checklist:


contracts/src/xYield/xYieldVault.sol
```diff
@@ -157,26 +161,26 @@ contract xYieldVault is ERC4626, Ownable2Step, ReentrancyGuard, Pausable {
...
+    function withdrawFrom(uint256 assets, address owner, uint64 chainId) external nonReentrant returns (uint256) {
+        _withdrawFrom(owner, assets);
+    }
+
+    // NOTE - for remote withdrawals we update virtualTotalAssets after withdrawal to prevent share price inflation
+    // before
+    // global shares is updated
+    function _withdrawFrom(address owner, uint256 assets) internal {
+        // TODO - deposit into escrow instead of transferring user's underlying to this contract
+        uint256 shares = yieldProtocol.withdraw(assets, address(this), address(this));
```
aharvet, line 166: If I understand well, we plan on potentially integrate with other lending protocols than Morpho and Euler ?
We need to keep in mind that not all of them uses erc 4624 and then they might not return shares here.

contracts/src/xYield/xYieldVault.sol
```diff
@@ -157,26 +161,26 @@ contract xYieldVault is ERC4626, Ownable2Step, ReentrancyGuard, Pausable {
...
+        emit WithdrawRemote(_msgSender(), owner, assets, shares);
```
aharvet: We can use `msg.sender` directly as eip 2771 is dead

contracts/src/xYield/xYieldVault.sol
```diff
@@ -17,13 +17,18 @@ contract xYieldVault is ERC4626, Ownable2Step, ReentrancyGuard, Pausable {
...
+    uint256 public minRebalanceGap;
+    uint256 public lastRebalanceTime;
     uint256 public virtualTotalSupply;
     bool public isActiveChain;
 
-    mapping(uint64 => address) public siblingVaults;
+    struct BalanceUpdate {
+        address recipient;
+        uint256 amount;
+        bool isMint;
```
aharvet: Semantically I would rather use an enum here to clearly stipulate if it's a deposit or a withdrawal. But not a big deal.

aharvet (changes requested):

contracts/src/xYield/xYieldVault.sol
```diff
@@ -157,26 +161,26 @@ contract xYieldVault is ERC4626, Ownable2Step, ReentrancyGuard, Pausable {
...
-    // }
+    function withdrawFrom(uint256 assets, address owner, uint64 chainId) external nonReentrant returns (uint256) {
+        _withdrawFrom(owner, assets);
+    }
+
+    // NOTE - for remote withdrawals we update virtualTotalAssets after withdrawal to prevent share price inflation
+    // before
+    // global shares is updated
+    function _withdrawFrom(address owner, uint256 assets) internal {
+        // TODO - deposit into escrow instead of transferring user's underlying to this contract
```
aharvet, line 165: Aren't we supposed to bridge it to caller's chain ?

evchip: correct. we do that via 7683, i.e. locking the user's funds in escrow until the solver fills it on the target chain + provides proof of fill

contracts/src/xYield/xYieldVault.sol
evchip: we will only integrate with 4626 vaults

contracts/src/xYield/xYieldVault.sol
evchip: fixed, thanks

contracts/src/xYield/xYieldVault.sol
evchip: changed, thanks

contracts/src/xYield/xYieldVault.sol
aharvet: But aren't we supposed to lock those funds on T1ERC7683 by calling `open` function ? or you want to implement custom logic to create an intent and settle after fill directly on xyield contract ?

evchip: yes, we want to open an intent. there are some edge cases around simply calling `open` within `_withdrawFrom`. What happens if the intent is not fulfilled and we need to initiate a refund? we would probably have to add functionality to claim the refund from the xYield contract. I'm still thinking about this, hence why it's marked TODO.

aharvet: Got it

aharvet (approved):
LGTM

PR #171, opened 2025-08-28
fix: changed type of orderId in signature to what the contract expects

kss-t1:
## Description

Another attempt at resolving a signature mismatch between Auction backend and T1ERC7683 contract.

Fix is :
- send `orderId` as a `bytes32` hex string rather than BigInt - which is what the contract expects.
- adjust `chainId` in domain header to be target chain id (was source)
- adjust `verifyingContractAddress` in domain header to be target chain contract (was source)

## How Has This Been Tested?

on v07

-   [ ] Docs

## Checklist:


aharvet (approved):
LGTM

PR #176, opened 2025-09-05
Feat/xyield/use across to bridge when rebalancing

kss-t1:
## Description

Changes T1ERC7683 to Across when executing `rebalance` .

## Motivation and Context

We need it because it would be tough to provide `mainnet` liquidity for `rebalance` to work - as of now. Therefore we use `Across` which is a large liquidity aggregator, with intentions to replace it with `T1ERC7683` in the future.

## How Has This Been Tested?

Unit tests with mocked Across contract.

-   [ ] Docs

## Checklist:


contracts/src/xYield/xYieldVault.sol
```diff
@@ -296,35 +304,42 @@ contract xYieldVault is ERC4626, Ownable2Step, ReentrancyGuard, Pausable {
...
+            asset(),
+            siblingVault.underlyingErc20,
+            _amountIn,
+            _amountOut,
+            _targetChain,
+            address(0),
+            _acrossApiQuoteTimestamp,
+            uint32(block.timestamp + 2 minutes),
+            0,
+            ""
```
evchip: if we pass calldata into this `message` param and implement [handleV3AcrossMessage](https://github.com/across-protocol/contracts/blob/5a4567c3e6d8908b1e0f9a0fc7d1f4c40fda5032/contracts/SpokePool.sol#L32) on this contract, it will send the message to the destination xYieldVault. This would be useful to emit an event like `RebalanceReceived`, which we can use offchain

evchip: and ideally this calldata should include a unique ID for the rebalance event for us to track.

evchip: I would also like to feed this unique ID as a param to this function. It would make my offchain life easier. It should be emitted from `Rebalanced` (probably should be named `RebalanceInitiated`)

evchip: I tackled this in [my PR](https://github.com/t1protocol/t1/pull/178) that would merge into this branch.

kss-t1: amazing

contracts/src/xYield/xYieldVault.sol
```diff
@@ -300,55 +310,49 @@ contract xYieldVault is ERC4626, Ownable2Step, ReentrancyGuard, Pausable {
...
-        virtualTotalAssets = ERC20(asset()).balanceOf(address(this));
-
-        (OrderData memory orderData) = abi.decode(order.orderData, (OrderData));
-        yieldProtocol.deposit(orderData.amountIn, address(this));
-
-        emit RebalanceUndone(orderData.destinationDomain, orderData.amountIn);
+        bytes memory message = abi.encode(_id);
+
+        acrossSpokePool.depositV3(
+            guardian,
```
evchip: On second thought, we should refund back to this vault. This is more transparent than sending funds back to the guardian's EOA

contracts/src/xYield/xYieldVault.sol
```diff
@@ -287,8 +296,9 @@ contract xYieldVault is ERC4626, Ownable2Step, ReentrancyGuard, Pausable {
         _setActiveChain(_isActive);
     }
 
-    function setSiblingVault(uint64 _chainId, address _vault) external onlyOwner {
-        siblingVaults[_chainId] = _vault;
+    function setSiblingVault(uint64 _chainId, address _vault, address _underlyingErc20) external onlyOwner {
+        siblingVaults[_chainId].vault = _vault;
+        siblingVaults[_chainId].underlyingErc20 = _underlyingErc20;
```
evchip, line 304: probably isn't a huge risk, but if we accidentally pass a zero address for _underlyingErc20, perhaps the solver still executes the request and effectively burns all the funds on fill. would be good to check it's not a zero address before setting.

evchip (dismissed):
just one nit. otherwise LGTM

PR #177, opened 2025-09-05
Feat/remote withdrawal

aharvet:
## Description

Implement `withdrawFrom` function bridging.

## Motivation and Context

The feature need to transfer tokens on the home chain of user.

## How Has This Been Tested?

Added tests in xYield test suite.

-   [ ] Docs

## Checklist:


contracts/src/xYield/xYieldVault.sol
```diff
@@ -116,6 +123,12 @@ contract xYieldVault is ERC4626, Ownable2Step, ReentrancyGuard, Pausable {
         _depositFrom(msg.sender, _receiver, _amount, shares, _chainId);
     }
 
+    function reDepositIdleAssets() external {
```
evchip: we don't want this, but especially without a modifier to restrict permissions. there may be reasons why we don't want to reinvest idle assets, such as funds temporarily being available for remote withdrawal by a user. anyway we have a logic determine to only invest funds if the chain is active, and this breaks that logic.

aharvet: This is meant to handle the case where across refunds the funds because fillDeadline passed after a withdrawFrom call.
How would you suggest to handle this case then ? Is gate keeping the function and letting the non evm app choosing if it should call it or not sufficient ?

evchip: You can gate it with `onlyGuardian`

aharvet: Done

evchip (approved):

PR #184, opened 2025-09-12
Fix/xyield total assets and shares update logic

evchip:
### `xYieldVault.sol`
- remove superfluous `virtualTotalAssets` variable
- increment/decrement `virtualTotalSupply` when remote deposit/withdraw
- use `previewShares` to determine shares to burn in `withdrawFrom`. we previously used the shares from the yield protocol's `withdraw` method, which would give us a totally incorrect value
- add is not active chain check on `updateTotals`, as we want to prevent updating totals on the active chain since it will already have correct total shares and shares minted
### xYield tests
- fix incorrect testing logic to use totalShares from the active chain to calculate the total global shates, rather than adding remote shares to totalShares, since totalShares from the active chain already accounts for global shares
- change virtualTotalSupply to totalSupply for simplicity
- add `testProveERC4626SameSharesMinted` to check that shares are correct before calling `updateTotals`

## Motivation and Context
virtualTotalSupply should have been updated inside _depositFrom and _withdrawFrom. The share price is determined by the total supply of share tokens and the total assets (USDC). The only thing that increases the share price is when the vault accrues yield. the asset quantity goes up, but the shares do not. so it’s important to not increase the share price for any other reason. when we deposit/withdraw (of any kind, including remote), we update the total assets. but in the _depositFrom and _withdrawFrom flow we were not updating the total shares. this is a big issue because it breaks the invariant.

We would subsequently increase/decrease the total shares when calling updateTotals on all chains, but this would be too late. It must be atomic to not break the share price. so instead, we only call updateTotals on the non active chain, hence the new check in updateTotals that the chain is not active.

## How Has This Been Tested?
Added a test and fixed existing tests.

## Checklist:

### DIFF
--- a/contracts/src/xYield/xYieldVault.sol
+++ b/contracts/src/xYield/xYieldVault.sol
@@ -25,7 +25,6 @@ contract xYieldVault is ERC4626, Ownable2Step, ReentrancyGuard, Pausable {
     IERC4626 public yieldProtocol;
     V3SpokePoolInterface public acrossSpokePool;
 
-    uint256 public virtualTotalAssets;
     uint256 public virtualTotalSupply;
     bool public isActiveChain;
     uint256 public rebalanceFillTTL = 2 minutes;
@@ -53,6 +52,7 @@ contract xYieldVault is ERC4626, Ownable2Step, ReentrancyGuard, Pausable {
     error InvalidChain();
     error WithdrawOnInactiveChain();
     error DepositOnInactiveChain();
+    error OnlyRemote();
     error NotImplemented();
     error LengthMismatch();
     error InvalidOrderData();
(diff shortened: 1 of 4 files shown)


contracts/src/xYield/xYieldVault.sol
```diff
@@ -229,7 +222,9 @@ contract xYieldVault is ERC4626, Ownable2Step, ReentrancyGuard, Pausable {
         internal
         returns (uint256 shares)
     {
-        shares = yieldProtocol.withdraw(assets, address(this), address(this));
+        shares = previewWithdraw(assets);
+        virtualTotalSupply -= shares;
```
kss-t1, line 226: By doing subtractions and additions of `virtualTotalSupply` manually, we may end up with values that are desynced from ones in `yieldProtocol` .

Is there a way to extract `virtualTotalSupply` from `yieldProtocol` instead? You seem to have beein doing this:
``` shares = yieldProtocol.withdraw(assets, address(this), address(this)); ``` , but now it's changed to manual calculation. Why?

evchip: yield protocol total supply != xYield virtual total supply. they account for different assets. the former is the Euler share token, the latter is the xYield share token. This line fixes conflating the two.

kss-t1 (approved):

PR #194, opened 2025-09-22
Feat/intent pauser

kss-t1:
## Description

Simple offchain application which monitors current USD value of `T1ERC7683` balance and pauses opening new intents if that balance is above certain `threshold` .

If `T1ERC7683` balance falls back under an acceptable value, it unpauses the contract.

It uses `Coingecko` API to get current ETH price for value at risk calculation.

## Motivation and Context

This is to limit money t1 (or Solvers) would lose in case our `T1ERC7683` contract gets exploited and an attacker finds a way to drain its funds.

-   [ ] Docs

## Checklist:


rymcol (approved):
Approved with comment: should we add a systemd service/monitor (ticket this?)

kss-t1:
> Approved with comment: should we add a systemd service/monitor (ticket this?)

I will do it as a part of this ticket (it's encapsulated as AC)

PR #200, opened 2025-09-26
Feat/7683/batch filler repayment

kss-t1:
## Description

Batch repayments for the 7683 intent fillers to optimize their gas costs.
- allows to process `Proofs of Read` in a batch
- when processing a batch of 7683 repayments, does some smart tricks to minimalise number of `_transferTokenOut` calls

Also introduces generic batch proof verification to `xChainRead`.

Functionality of processing a PoRs one by one was unchanged.

## Motivation and Context

We want to beat bridging prices of our competition, so we want to show best profitability for solvers. Hence the gas optimization.

## How Has This Been Tested?

Unit test which confirms correctness and gas savings. In the unit test, a simple [Merkle Trie](https://github.com/dmfxyz/murky) implementation was used to create Merkle Trie. 

-   [ ] Docs

## Checklist:


evchip:
@kss-t1 `forge fmt` please

evchip:
> Not tested.

y no u unit test?

contracts/src/7683/T1ERC7683.sol
```diff
@@ -419,15 +424,80 @@ contract T1ERC7683 is IT1ERC7683, T1Permit2, AccessControlUpgradeable, EIP712 {
...
+        for (uint256 i = 0; i < encodedProofsOfRead.length; i++) {
+            (bytes32 orderId, bool isSettled, address inputToken, address settlementReceiver, uint256 amount) =
+                _decodeOrdersToSettle(requestIds[i], results[i]);
+
+            orderIds[i] = orderId;
+            areSettled[i] = isSettled;
+
+            if (isSettled) {
+                bool foundExistingUserTokenKey = false;
+                for (uint256 j = 0; j < uniqueUserTokenCount; j++) {
```
evchip: uniqueUserTokenCount is initialized to zero. So this loop won't run

contracts/src/7683/T1ERC7683.sol
```diff
@@ -419,15 +424,80 @@ contract T1ERC7683 is IT1ERC7683, T1Permit2, AccessControlUpgradeable, EIP712 {
...
+                        foundExistingUserTokenKey = true;
+                        break;
+                    }
+                }
+
+                if (!foundExistingUserTokenKey) {
+                    userTokenKeys[uniqueUserTokenCount] = UserTokens({ receiver: settlementReceiver, token: inputToken });
+                    amounts[uniqueUserTokenCount] = amount;
+                    uniqueUserTokenCount++;
+                }
```
evchip: this is super not readable for me and if I understand correctly it's just a gas optimization to aggregate multiple transfers to the same solver into one. Not worth the smoke IMO. I wouldn't be surprised if it's actually more gas this way.

contracts/src/7683/T1ERC7683.sol
```diff
@@ -419,15 +424,80 @@ contract T1ERC7683 is IT1ERC7683, T1Permit2, AccessControlUpgradeable, EIP712 {
...
         if (orderId == bytes32(0)) revert InvalidOrder();
 
         delete settlementReadRequestToOrderId[requestId];
 
         // Check if the order is FILLED based on result length
-        bool isSettled = (result.length != 0);
+        isSettled = (result.length != 0);
+        inputToken = address(0);
+        settlementReceiver = address(0);
+        amount = 0;
```
evchip: this is redundant, since it's just setting all of the defaults. you'll get the same result by deleting these 4 lines.

evchip (changes requested):
Found a few issues on first pass.

contracts/src/7683/T1ERC7683.sol
kss-t1: It won't run on a first try only, which is expected. Will explain in person.

contracts/src/7683/T1ERC7683.sol
kss-t1: I think it will cost A LITTLE more gas in a scenario where every proof in a batch is for a different <EOA,token> pair. For vast majority of scenarios it will save quite a lot of gas, I think.


Let's spec it as you suggested.

contracts/src/7683/T1ERC7683.sol
evchip: Ok, perhaps I missed something. It's very hard to read this level of nesting.

evchip: Yes, I see now. Thanks for extracting into a function

contracts/src/test/utils/MurkyMerkleBase.sol
```diff
@@ -0,0 +1,203 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+abstract contract MurkyMerkleBase {
```
evchip, line 4: could we not have added this as a lib?

evchip (approved):

contracts/src/test/utils/MurkyMerkleBase.sol
kss-t1: We could, but it would require importing a gitmodule. That way it is much lighter for the build and repo.

PR #222, opened 2025-12-18
Refactor: improve sealed bid auction server recovery

evchip:
## Description
Overall just adding more robustness and defensive programming to the sealed-bid-auction app:
- add try/catch block to auction call
- add optional "from" blocks as env vars for historical event logging. these are useful if we miss events and want to rewind
- send `Open` event metadata to solvers so that they do not have to re-fetch this event onchain themselves. This was unnecessary complexity that made it difficult to replay missed open intent events.

## Motivation and Context
the sealed bid auction server experienced a failure fetching arbitrum `Open` intent events that became difficult to recover from. These changes **do not** contribute to making these kind of failures less likely, but they do help us to more easily manually recover from a failure

## How Has This Been Tested?
Yes, manually tested in prod.

## Checklist:


contracts/deployments/7683/v0.9.1.json
evchip, line 1: idk why gh is showing this as a diff. it's already been merged.

tee-apps/sealed-bid-auction/.env.template
```diff
@@ -7,6 +7,11 @@ SOLVER_PRICE_TTL_SECONDS=600
 NEXT_NONCE_URL=
 IS_MAINNET=false
 
+# optional from blocks for historical event logging
+# these are useful if we miss events and want to rewind
+INTENT_OBSERVER_FROM_BLOCK_BASE=
```
kss-t1, line 12: Intents have `fillDeadline` . Normally this deadline is not going to be far into the future. Therefore, if we missed `Open` events and want to rewind, these events likely will have their `fillDeadline` already expired.

Therefore it seems like rewinding past events is not very useful. In case of `Sealed Bid Auction Server` downtime, users should be in my opinion expected to refund/cancel their expired Intents (as they should anyway - if Intents expire for a different reason (eg. no Solver available) !) .

Additionally - in this implementation we don't seem to protect ourselves from processing past intents in the block range that already have been processed. Consequently, we might end up running an auction for an Intent that already has been solved. Perhaps such case could be handled in the Solver implementation (eg. Solver could just skip intents that were already filled) but I don't see a reason that would make it worth it.

Therefore - unless we have a good reason to  to merge rewinding missed Intents in, I don't think we should.

kss-t1 (changes requested):

kss-t1:
Asked for changes because I am not a fan of rewinding missed Intents.

Rest of the changes look good.

tee-apps/sealed-bid-auction/.env.template
evchip: fill deadline seems to be some time between 24 and 48 hours in the future. I still think it's useful to rewind for these rather than let them fall into refund status.

> Additionally - in this implementation we don't seem to protect ourselves from processing past intents in the block range that already have been processed.

That's true but it's really no harm other than a bit of wasted gas for creating another `fill` tx that reverts because it was already filled.

kss-t1 (approved):
Approved after consulting - rewinding might be useful, and the Filler is able to handle double processing of an intent gracefully.

PR #224, opened 2025-12-31
Feat/sealed bid auction/ws outage resistance

kss-t1:
## Description

Makes our `Sealed Bid Auction Server` resistant to WebSocket connection dropping due to external factors by:
- persistently reconnecting after connection drops
- processing missed blocks after WS reconnects 

Also, removes `await` on processing a single intent auction, so that next intents are processed in parallel. Without that, if no solver is found for an intent then processing other intents is blocked until we reach `fillDeadline` for that intents (which could be a long time) .

## Motivation and Context

Because we only tried to reconnect 5 times when WS connection dropped, which resulted in no longer listening to inbound intents. Also, we were missing intents every now and then if they happened when WS connection was down.

## How Has This Been Tested?

tested on local, against mainnet contracts!

-   [ ] Docs

## Checklist:

### DIFF
--- a/tee-apps/sealed-bid-auction/bun.lock
+++ b/tee-apps/sealed-bid-auction/bun.lock
@@ -1,16 +1,17 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "sealed-bid-auction",
       "dependencies": {
         "@jest/globals": "^30.0.5",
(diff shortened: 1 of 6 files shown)


evchip (approved):
Looks good to me! Only thing I think could be improved is a PagerDuty hook. But I have that in my open PR. Can rework it once we merge this in

